Issue #9, opened 2018-09-24
Prettify tables


maelle:
Right now, the number of dashes I put under the header is 3 no matter what's in the table.

It'd be nice to loop through all rows to find the longest element for each column, and decide the number of dashes based on that. Probably more XSLT than I ever wished to learn, but would be great.

maelle:
Anyone with XSLT skills who'd be willing to help improve the XSLT stylesheet is welcome to contribute.

pdaengeli:
I ended up here by accident, but let's see if I can help.

Not sure I understand the exact requirements. As it seems you are using xsltproc and thus tied to XSLT 1.0.

Anyways, here is one way you could approach this:

```xml
    <!-- Table -->
    
    <xsl:template match="md:table">
        <xsl:value-of select="$maxLength"/>
        <xsl:apply-templates select="." mode="indent-block"/>
        <xsl:apply-templates select="md:*"/>
    </xsl:template>
    
    <xsl:variable name="maxLength">
        <xsl:for-each select="//md:table_header/md:table_cell">
            <xsl:variable name="pos" select="position()"/>
            <!-- EXslt or XSLT 1.1 would be needed to lookup node-sets; 
                thus generating a string (something like CELL1:7|CELL2:5|CELL3:9|CELL4:8|) -->
            <xsl:text>CELL</xsl:text>
            <xsl:value-of select="$pos"/>
            <xsl:text>:</xsl:text>
            <xsl:for-each select="//md:table_cell[position()=$pos]/md:text">
                <xsl:sort data-type="number" select="string-length()" order="descending"/>
                <xsl:if test="position()=1">
                    <xsl:value-of select="string-length()"/>
                    <xsl:value-of select="'|'"/>
                </xsl:if>
            </xsl:for-each>
        </xsl:for-each>
    </xsl:variable>
    
    <!-- recursive template to print n dashes/characters -->
    <xsl:template name="n-times">
        <xsl:param name="n"/>
        <xsl:param name="char"/>
        <xsl:if test="$n > 0">
            <xsl:call-template name="n-times">
                <xsl:with-param name="n" select="$n - 1"/>
                <xsl:with-param name="char" select="$char"/>
            </xsl:call-template>
            <xsl:value-of select="$char"/>
        </xsl:if>
    </xsl:template>
    
    <xsl:template match="md:table_header">
        <xsl:text>| </xsl:text>
        <xsl:apply-templates select="md:*"/>
[...]

PR #13, opened 2020-05-12
add sourcepos argument

zkamvar:
## Description

I've added  a `sourcepos` argument to `to_xml()`; this allows the source position to be added as an attribute to the XML. It defaults to FALSE. 

## Example

``` r
library("tinkr")
path <- system.file("extdata", "example1.md", package = "tinkr")
post_list <- to_xml(path, sourcepos = TRUE)
xml2::xml_attr(post_list[[2]], "sourcepos")
#> [1] "1:1-446:63"
```

<sup>Created on 2020-05-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,7 +20,7 @@ Description: Casts (R)Markdown files to XML and back to allow their editing via 
 License: GPL-3
 Encoding: UTF-8
 LazyData: true
-RoxygenNote: 6.1.0.9000
+RoxygenNote: 7.1.0
 Imports: 
     magrittr,
     commonmark (>= 1.6),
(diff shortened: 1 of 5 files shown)


zkamvar:
I may be wrong, but the sourcepos attribute could possibly be used to fix the tables issue by looping over the column range?

maelle:
Thank you! To be honest, it's been a while since I looked at this but you're right, it might help with #9 

Do you use tinkr for anything in particular?

zkamvar:
At the moment, I'm working on https://github.com/zkamvar/up2code to get ready for updating the Carpentries lesson template. I use tinkr to read in the data and compare the output when I make broad level changes. It's really handy :)

maelle:
I had to go back and read my own tech note to really remember how `tinkr` works and I got excited all over again :joy: It is very cool that you found an use case for it, and that's a very interesting use case!

PR #14, opened 2020-05-15
ensure Rmd chunks retain sourcepos attribute

zkamvar:
I noticed that the `sourcepos` attribute was being stripped when the chunk info was being added. It seems that xml2 will strip attributes when new attributes are added. I've fixed it in this PR

``` r
library(magrittr)
tinkr::to_xml(path, sourcepos = TRUE)$body %>%
  xml2::xml_find_first(".//d1:code_block") %>%
  xml2::xml_attrs()
#>  sourcepos      space   language       name    include       eval 
#>  "2:1-4:3" "preserve"        "r"    "setup"    "FALSE"     "TRUE"
```

<sup>Created on 2020-05-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/to_xml.R
+++ b/R/to_xml.R
@@ -57,7 +57,12 @@ transform_block <- function(code_block){
   info <- transform_params(info)
 
   xml2::xml_set_attr(code_block, "info", NULL)
-  xml2::xml_set_attrs(code_block, info)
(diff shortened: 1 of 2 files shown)


maelle:
Thank you!

PR #15, opened 2020-05-28
prevent to_md() from clobbering document (and friends!)

zkamvar:
## Description

I've done a couple of things here (I apologize for shoving too many things in a single PR, but I figured they were all roughly related):

1. Fixed a bug I introduced in #13/#14 where code chunks would have an extra `sourcepos` option added if read in with sourcepos = TRUE (whoops :flushed:)
2. Updated `to_xml()` to only post-process code blocks that have a language attribute to allow bare or unevaluated code fences.
3. I noticed that the original xml document was being affected by the round trip (because xml2 objects operate on pass-by-reference, which is summarized SO WELL in [this coffee cup gif](https://blog.penjee.com/wp-content/uploads/2015/02/pass-by-reference-vs-pass-by-value-animation.gif)), so I have updated the processing to copy the document before processing and it no longer goes through a disk write (there was a weird namespace issue with this process that was affecting the block processing, so I use [the technique I found in my own package](https://github.com/zkamvar/up2code/blob/2a8bc3072b9db49777ad4ab4c6eef76bf3a85537/R/Episode.R#L273-L286))

## Example

```` r
library(tinkr)
library(magrittr)
rmd <- tinkr::to_xml(path, sourcepos = TRUE)
rmd$body %>%
  xml2::xml_find_first(".//d1:code_block") %>%
  xml2::xml_attrs()
#>  sourcepos      space   language       name    include       eval 
#>  "2:1-4:3" "preserve"        "r"    "setup"    "FALSE"     "TRUE"
tmp <- tempfile()
to_md(rmd, tmp)
# blocks are processed correctly
readLines(tmp) %>%
  head(10) %>%
  cat(sep = "\n")
#> ---
#> title: "Untitled"
#> author: "M. Salmon"
#> date: "September 6, 2018"
#> output: html_document
#> ---
#> 
#> ```{r setup, include=FALSE, eval=TRUE}
#> knitr::opts_chunk$set(echo = TRUE)
#> ```
# unevaluated blocks are processed correctly
readLines(tmp) %>%
  tail(29) %>%
  head(10) %>%
  cat(sep = "\n")
#> 
#> Non-RMarkdown blocks are also considered
#> 
#> ```bash
#> echo "this is an unevaluted bash block"
#> ```
#> 
#> ```
#> This is an ambiguous code block
#> ```
# The attributes have not changed
rmd$body %>%
  xml2::xml_find_first(".//d1:code_block") %>%
  xml2::xml_attrs()
#>  sourcepos      space   language       name    include       eval 
#>  "2:1-4:3" "preserve"        "r"    "setup"    "FALSE"     "TRUE"
````

<sup>Created on 2020-05-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### CODE REVIEW

File: R/to_md.R [hunk 1]
```diff
@@ -26,56 +26,68 @@
 #' xml2::xml_set_attr(headers3, "level", 1)
 #' yaml_xml_list$body <- body
 #' # save back and have a look
-#' to_md(yaml_xml_list, "newmd.md")
+#' newmd <- tempfile("newmd", fileext = ".md")
+#' to_md(yaml_xml_list, newmd)
 #' # file.edit("newmd.md")
+#' file.remove(newmd)
```
maelle, line 32: much better! should I submit this to CRAN one day? If I do you'll have to become an author.

R/to_md.R
```diff
@@ -26,56 +26,68 @@
...
+  xml2::xml_set_attr(new_text, dff, NULL)
+
+  new
 }
 
 transform_code_blocks <- function(xml){
+  # Find all code blocks with a language attribute (those without it are not processed)
   code_blocks <- xml %>%
-    xml2::xml_find_all(xpath = './/d1:code_block',
+    xml2::xml_find_all(xpath = './/d1:code_block[@language]',
```
maelle, line 75: so much more elegant! :sparkles:

maelle:
Would it be fine to add you as an author before merging? If so, please go ahead and add your metadata.

R/to_md.R
zkamvar: It would be nice to have on CRAN at some point, but I might like to give #9 a shot first :wink:

PR #20, opened 2020-09-22
checkboxes

maelle:
At the moment checkboxes are lost (escaped) so we'll need to tweak the XSLT stylesheet.

### DIFF
--- a/inst/extdata/checkboxes.md
+++ b/inst/extdata/checkboxes.md
@@ -1,0 +1,8 @@
+---
+title: "Checkboxes"
+---
+
+- [ ] blabla
+
+- [x] blabla
+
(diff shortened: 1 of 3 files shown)


zkamvar:
We had this issue in {pegboard} with links that were included from a separate document (e.g. there would be a link like [download RStudio][rstudio], where [rstudio] was defined in links.md). Our solution was:

1. find the text that are valid links 
2. add a tag to their element (we called it 'klink')
3. specially process those tags as unescaped text: https://github.com/carpentries/pegboard/blob/03fb639993960115aec6e8bf4cd3b45b200efdf1/inst/stylesheets/xml2md_gfm_kramdown.xsl#L22-L26

PR #24, opened 2020-09-22
only set info attribute for language blocks

### LINKED ISSUE #23: Question: how do you insert new code chunks?
I've been thinking about the proper way to go about this and It's still hard for me to wrap my brain around. One use-case I have is that I want to be able to insert a setup chunk in a whole bunch of RMarkdown documents that don't have them.

I'm running into a problem where the chunk gets excluded from the document. 

```r
x <- tinkr::to_xml(system.file("extdata", "example2.Rmd", package = "tinkr"))

make_code_block <- function(code = "print('hello world?')", name = 'cody', body) {
  block <- glue::glue("
    <document xmlns='{xml2::xml_ns(body)[[1]]}'>
      <code_block language='r' name='{name}' echo='TRUE' eval='TRUE'>
        {code}
      </code_block>
    </document>"
  )
  xml2::xml_child(xml2::read_xml(block))
}

# insert a new code block after the setup chunk
xml2::xml_add_child(x$body, make_code_block(body = x$body), .where = 1)

# find that code block
xml2::xml_find_all(x$body, ".//d1:code_block[@name='setup']")
#> {xml_nodeset (1)}
#> [1] <code_block xml:space="preserve" language="r" name="setup" include="FALSE ...
xml2::xml_find_all(x$body, ".//d1:code_block[@name='cody']")
#> {xml_nodeset (1)}
#> [1] <code_block xmlns="http://commonmark.org/xml/1.0" language="r" name="cody ...

# Round trip does not work
tou <- tempfile(fileext = ".Rmd")
tinkr::to_md(x, tou)
xml2::xml_find_all(tinkr::to_xml(tou)$body, ".//d1:code_block[@name='setup']")
#> {xml_nodeset (1)}
#> [1] <code_block xml:space="preserve" language="r" name="setup" include="FALSE ...
xml2::xml_find_all(tinkr::to_xml(tou)$body, ".//d1:code_block[@name='cody']")
#> {xml_nodeset (0)}
```

<sup>Created on 2020-09-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I'm noticing that the namespace attribute is being swept up into the code block and it may be beneficial to add it to the list of tags to avoid: https://github.com/ropenscilabs/tinkr/blob/4980e816d876a86bb1ea07fe2b564701bc4bce22/R/to_md.R#L93

zkamvar:
This will fix #23 via https://github.com/ropenscilabs/tinkr/pull/24/files#diff-cf43f9aae06899fecb008822d7d986c4R113, which sets the "info" attribute instead of resetting ALL attributes. The only potential drawback is if any of the chunk-derived attributes interact with the stylesheet, but I can't see any pathways so far.

I've also added some documentation in the README about the process for adding new elements (which really should be a vignette on its own):

````r
yaml_xml_list <- tinkr::to_xml(path)

# Namespace comes from commonmark
xml2::xml_ns(yaml_xml_list$body)
#> d1 <-> http://commonmark.org/xml/1.0

new_chunk <- xml2::read_xml("<code_block language='r' name='xml-block'></code_block>")

# set the namespace of the new chunk
xml2::xml_set_attr(new_chunk, "xmlns", xml2::xml_ns(yaml_xml_list$body))

# add code in the new chunk
# NOTE: code MUST end in a newline character
xml2::xml_set_text(new_chunk, "cat('this is a new chunk from {tinkr}')\n")
#> {xml_document}
#> <code_block language="r" name="xml-block" xmlns="http://commonmark.org/xml/1.0">

# Add chunk into document
xml2::xml_add_child(yaml_xml_list$body, new_chunk, .where = 1L)

newfile <- tempfile(fileext = ".Rmd")
tinkr::to_md(yaml_xml_list, newfile)
cat(readLines(newfile, 16), sep = "\n")
#> ---
#> title: "Untitled"
#> author: "M. Salmon"
#> date: "September 6, 2018"
#> output: html_document
#> ---
#> 
#> ```{r setup, include=FALSE, eval=TRUE}
#> knitr::opts_chunk$set(echo = TRUE)
#> ```
#> 
#> ```{r xml-block}
#> cat('this is a new chunk from {tinkr}')
#> ```
#> 
#> ## R Markdown
````

### DIFF
--- a/R/to_md.R
+++ b/R/to_md.R
@@ -90,7 +90,7 @@ transform_code_blocks <- function(xml){
 to_info <- function(code_block){
  attrs <- xml2::xml_attrs(code_block)
  options <- attrs[!names(attrs) %in%
-                  c("language", "name", "space", "sourcepos")]
+                  c("language", "name", "space", "sourcepos", "xmlns")]
 
  if(length(options) > 0){
    options <- glue::glue("{names(options)}={options}") %>%
@@ -101,6 +101,7 @@ to_info <- function(code_block){
  }
 
  if (attrs["name"] != ""){
+   # if (attrs["name"] == "cody") browser()
    attrs["name"] <- paste0(" ", attrs["name"])
  }
 
@@ -109,5 +110,5 @@ to_info <- function(code_block){
  info <- paste0(info, "}")
  names(info) <- "info"
 
- xml2::xml_set_attrs(code_block, info)
+ xml2::xml_set_attr(code_block, "info", info)
 }
--- a/R/to_md.R
+++ b/R/to_md.R
@@ -101,7 +101,6 @@ to_info <- function(code_block){
  }
 
  if (attrs["name"] != ""){
-   # if (attrs["name"] == "cody") browser()
    attrs["name"] <- paste0(" ", attrs["name"])
  }
 
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ to_md(yaml_xml_list, "newmd.md")
 file.edit("newmd.md")
 ```
 
-### RMarkdown
+### R Markdown
 
 For R Markdown files, to ease editing of chunk label and options, `to_xml` munges the chunk info into different attributes. E.g. below you see that `code_blocks` can have a `language`, `name`, `echo` attributes.
 
--- a/README.md
+++ b/README.md
@@ -95,47 +95,41 @@ say we wanted to add a new R code chunk after the setup chunk:
 > NOTE: Inserting new code MUST have a newline character at the end of the
 > chunk or else the last line will be lost.
 
-````r
+``` r
 path <- system.file("extdata", "example2.Rmd", package = "tinkr")
 yaml_xml_list <- tinkr::to_xml(path)
-
-# Namespace comes from commonmark-  xml2::xml_ns(yaml_xml_list$body)
-#> d1 <-> http://commonmark.org/xml/1.0
-
-new_chunk <- xml2::read_xml("<code_block language='r' name='xml-block'></code_block>")
-
-# set the namespace of the new chunk
-xml2::xml_set_attr(new_chunk, "xmlns", xml2::xml_ns(yaml_xml_list$body))
-
-# add code in the new chunk
-# NOTE: code MUST end in a newline character
-xml2::xml_set_text(new_chunk, "cat('this is a new chunk from {tinkr}')\n")
-#> {xml_document}
-#> <code_block language="r" name="xml-block" xmlns="http://commonmark.org/xml/1.0">
-
 # Add chunk into document
-xml2::xml_add_child(yaml_xml_list$body, new_chunk, .where = 1L)
+xml2::xml_add_child(yaml_xml_list$body, 
+                    "code_block",
+                    "message(\"this is a new chunk from {tinkr}\")\n",
+                    language='r', 
+                    name='xml-block', 
+                    xmlns=xml2::xml_ns(yaml_xml_list$body)[[1]],
+                    .where = 1L)
+out <- tempfile(fileext = ".Rmd")
+tinkr::to_md(yaml_xml_list, out)
+file.edit(out)
+```
 
-newfile <- tempfile(fileext = ".Rmd")
-tinkr::to_md(yaml_xml_list, newfile)
-cat(readLines(newfile, 16), sep = "\n")
-#> ---
-#> title: "Untitled"
-#> author: "M. Salmon"
-#> date: "September 6, 2018"
-#> output: html_document
-#> ---
-#> 
-#> ```{r setup, include=FALSE, eval=TRUE}
-#> knitr::opts_chunk$set(echo = TRUE)
-#> ```
-#> 
-#> ```{r xml-block}
-#> cat('this is a new chunk from {tinkr}')
-#> ```
-#> 
-#> ## R Markdown
+````markdown
+---
+title: "Untitled"
+author: "M. Salmon"
+date: "September 6, 2018"
+output: html_document
+---
+
+```{r setup, include=FALSE, eval=TRUE}
+knitr::opts_chunk$set(echo = TRUE)
+```
+
+```{r xml-block}
+message("this is a new chunk from {tinkr}")
+```
+
+## R Markdown
 ````
+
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: R/to_md.R [hunk 1]
```diff
@@ -101,6 +101,7 @@ to_info <- function(code_block){
  }
 
  if (attrs["name"] != ""){
+   # if (attrs["name"] == "cody") browser()
```
maelle, line 104: why is there commented code?

zkamvar: whoops! That was my debugging :flushed:

zkamvar, line 104:
```suggestion
   # if (attrs["name"] == "cody") browser()
```

README.md
```diff
@@ -62,6 +64,8 @@ to_md(yaml_xml_list, "newmd.md")
 file.edit("newmd.md")
 ```
 
+### RMarkdown
```
maelle, line 67:
```suggestion
### R Markdown
```

:nail_care:

R/to_md.R
zkamvar, line 104:
```suggestion
```

README.md
```diff
@@ -82,7 +86,56 @@ yaml_xml_list$body
 #> [10] <paragraph>\n  <text>Note that the </text>\n  <code>echo = FALSE</c ...
 ```
 
+### Inserting new elements
+
+You can insert new elements into the document via {xml2}, but you should make
+sure that the namespace matches that of your xml document. For example, let's
+say we wanted to add a new R code chunk after the setup chunk:
 
+> NOTE: Inserting new code MUST have a newline character at the end of the
```
maelle, line 95: why, btw?

zkamvar: I'm not sure! I think it has something to do with the stylesheet because it's accessible before it hits `xslt::xml_xslt()`. I think it may be worthwhile to open an issue to figure out what's going on, but probably not too high priority IMO.

tests/testthat/test-to_md.R
```diff
@@ -68,3 +68,42 @@ test_that("to_md does not break tables", {
...
   testthat::expect_snapshot_file(newmd)
 })
+
+test_that("code chunks can be inserted on round trip", {
+
+  path <- system.file("extdata", "example2.Rmd", package = "tinkr")
+  tmpdir <- tempdir()
+  suppressWarnings(dir.create(tmpdir))
+  newmd <- file.path(tmpdir, "ex.Rmd")
+  on.exit(file.remove(newmd))
```
maelle, line 78: should we use self-cleaning test fixtures? https://www.tidyverse.org/blog/2020/04/self-cleaning-test-fixtures/

maelle:
Not sure if helpful but I used your edit of `to_md()` and the code below

```r
yaml_xml_list <- tinkr::to_xml(path)
# Add chunk into document
xml2::xml_add_child(yaml_xml_list$body, 
                    "code_block",
                    "this is a new chunk from {tinkr}\n",
                    language='r', name='xml-block', 
                    xmlns=xml2::xml_ns(yaml_xml_list$body),
                    .where = 1L)
tou <- tempfile(fileext = ".Rmd")
tinkr::to_md(yaml_xml_list, tou)
file.edit(tou)
```

I was wondering whether it was more natural to add a chunk this way? I.e. not creating a node, but instead using the "..." of `xml2::xml_add_child()`. I can't remember where I saw that, I used GitHub code search to look at examples.

The namespace thing is tricky, should your PR actually add a helper function for adding childs (and siblings and parents?)? The arguments would be everything but not the namespace, that tinkr would set. The function would also check the text ends with a newline, if there's text.

I saw you're into R6 and I wonder whether that's the route we should go in tinkr too instead of using a list?

tests/testthat/test-to_md.R
maelle: in particular https://withr.r-lib.org/reference/with_file.html

zkamvar: +10000 
|
I'm definitely for the self-cleaning test fixtures, but I would prefer to convert them in a separate PR after this is merged.

zkamvar:
> I was wondering whether it was more natural to add a chunk this way? I.e. not creating a node, but instead using the "..." of `xml2::xml_add_child()`. I can't remember where I saw that, I used GitHub code search to look at examples.

YES! I think the method you proposed is a MUCH better way of doing things. I admit that I'm always a bit perplexed by the {xml2} documentation, and never really understood what was going on with the `...`, but now it makes sense!

> The namespace thing is tricky, should your PR actually add a helper function for adding childs (and siblings and parents?)? The arguments would be everything but not the namespace, that tinkr would set. The function would also check the text ends with a newline, if there's text.

I think adding helper functions for these situations would be quite nice. The only thing is that the code block example is one of the simplest situations because you only need the code block and the text that goes inside. It gets more complex with almost any other element. 

One avenue we might explore is having a helper function that inserts user markdown as nodes by passing it through `commonmark::markdown_xml()` and then taking the children of that new document. That's was one of the strategies I found to work with [dealing with list elements in {pegboard}](https://github.com/carpentries/pegboard/blob/f2155d3a5e6b94ff677fdf2daeed9562458e6085/R/utils.R#L106-L118)

maelle:
Well for now the perceived audience is you and me 🤣

zkamvar:
> It gets more complex with almost any other element.

This may not necessarily be the case AND I just realized that this could tie into #20 because if the intention is to allow users to insert any random markdown, then it can be inserted into a text node and it would come out in the rendered document as markdown. The only things we would need to do is to add a tag to that text that would prevent it from be escaped on its way out. I [used this strategy in {pegboard}](https://github.com/carpentries/pegboard/blob/1d0db8b3b3bc2b92953ee5009c299ce037b8b9a8/inst/stylesheets/xml2md_gfm_kramdown.xsl#L22-L26).

zkamvar:
@maelle, do you think this can be merged? This bugfix would really help me refactor {pegboard}.

tests/testthat/test-to_md.R
zkamvar: I opened #25 to track this

zkamvar:
Now that **<SNAP! voice> I've got the power </SNAP! voice>** to merge PRs, I will do this self-merge since I have addressed @maelle's requests.

PR #31, opened 2020-10-21
use simpler process for copying document

zkamvar:
I realized that it is simpler to convert the document to text and re-read it than to go through the whole "new_root" path (which I learned through https://community.rstudio.com/t/adding-nodes-in-xml2-how-to-avoid-duplicate-default-namespaces/84870/2).


maelle:
nice!!

Issue #28, opened 2020-10-08
Add helper functions for adding/extracting features


zkamvar:
Is it worthwhile to add helper functions for adding or extracting nodes? There is an issue around namespaces, which can be tricky. Things I think could be on the docket to add:

 - code chunks
 - lists
 - tables

> The namespace thing is tricky, should your PR actually add a helper function for adding childs (and siblings and parents?)? The arguments would be everything but not the namespace, that tinkr would set. The function would also check the text ends with a newline, if there's text.

I think adding helper functions for these situations would be quite nice. The only thing is that the code block example is one of the simplest situations because you only need the code block and the text that goes inside. It gets more complex with almost any other element. 

One avenue we might explore is having a helper function that inserts user markdown as nodes by passing it through `commonmark::markdown_xml()` and then taking the children of that new document. That's was one of the strategies I found to work with [dealing with list elements in {pegboard}](https://github.com/carpentries/pegboard/blob/f2155d3a5e6b94ff677fdf2daeed9562458e6085/R/utils.R#L106-L118)

_Originally posted by @zkamvar in https://github.com/ropensci/tinkr/pull/24#issuecomment-699036027_

zkamvar:
Note that this issue ties closely with #27

zkamvar:
After getting lost in namespace weeds a bit, it appears that this feature is definitely needed because there is no way to cleanly add nodes into the default namespace: https://community.rstudio.com/t/adding-nodes-in-xml2-how-to-avoid-duplicate-default-namespaces/84870/2

maelle:
now that one can add Markdown, is this still needed?

PR #33, opened 2020-11-04
Use R6 representation

### LINKED ISSUE #25: Use self-cleaning test fixtures
> should we use self-cleaning test fixtures? https://www.tidyverse.org/blog/2020/04/self-cleaning-test-fixtures/

Context: this package was developed before self-cleaning test fixtures via {withr} was a thing, but they definitely will help clean up the readability, so it's worth a small refactor :)

_Originally posted by @maelle in https://github.com/ropensci/tinkr/pull/24#discussion_r494794519_

### LINKED ISSUE #27: Use R6 to store the object?
This was brought up in a PR review:

> > I saw you're into R6 and I wonder whether that's the route we should go in tinkr too instead of using a list?
> 
> I'm ambivalent about this. I've found it to be quite useful for my own tasks and it reinforces the concept that {xml2} objects work by side-effect, but on the other hand, it may not be strictly necessary. I guess it depends on what the perceived audience will do with it. :shrug:

_Originally posted by @zkamvar in https://github.com/ropensci/tinkr/pull/24#issuecomment-699036027_

zkamvar:
I've added an R6 class object with four methods: new, write, reset, and add_md. 

I've also added test fixtures and snapshot tests (which are super fun!) (I also may have been on the rOpenSci allcaps when I was writing some of the tests...)

this PR will fix #25, fix #27, and address #28


zkamvar:
> I've added an R6 class object with four methods: new, write, reset, and add_md.
> 
> I've also added test fixtures and snapshot tests (which are super fun!)
> 
> this PR will fix #25, fix #27, and address #28

R/class-yarn.R
```diff
@@ -0,0 +1,134 @@
...
+    #' ex1 <- tinkr::yarn$new(path)
+    #' ex1
+    #' tmp <- tempfile()
+    #' try(readLines(tmp)) # nothing in the file
+    #' ex1$write(tmp)
+    #' head(readLines(tmp)) # now a markdown file
+    #' unlink(tmp)
+    write = function(path = NULL, 
+      stylesheet_path = system.file("extdata", "xml2md_gfm.xsl", package = "tinkr")){
+      if (is.null(path)) {
```
maelle, line 89: should we also check both paths exist?

R/class-yarn.R
```diff
@@ -0,0 +1,134 @@
...
+        stop("Please provide a file path", call. = FALSE)
+      }
+      to_md(self, path, stylesheet_path)
+      invisible(self)
+    },
+    
+    #' @description add an arbitrary markdown element to the document
```
maelle, line 96:
```suggestion
    #' @description add an arbitrary Markdown element to the document
```

R/class-yarn.R
```diff
@@ -0,0 +1,134 @@
...
+    #' @examples
+    #' path <- system.file("extdata", "example1.md", package = "tinkr")
+    #' ex1 <- tinkr::yarn$new(path)
+    #' ex1
+    #' tmp <- tempfile()
+    #' try(readLines(tmp)) # nothing in the file
+    #' ex1$write(tmp)
+    #' head(readLines(tmp)) # now a markdown file
+    #' unlink(tmp)
+    write = function(path = NULL,
```
maelle, line 87: should there also be a way to print to screen easily? I.e. not writing to a file but to the console, for a quick check or so.

maelle (approved):
This is awesome, thank you! I left a few comments.

Another question that I have is how this should be documented in the README. What approach would we recommend first, R6 or not? In any case at the very top of the README where there's a list of use cases you could add "add an arbitrary new Markdown element like a table".

R/class-yarn.R
```diff
@@ -0,0 +1,134 @@
...
+#' publicly accessible slots: path, yaml, body, and ns.
+#' @details
+#' This class is a fancy wrapper around the results of [tinkr::to_xml()] and
+#' has methods that make it easier to add, analyze, remove, or write elements
+#' of your markdown document.
+#' @export
+yarn <- R6::R6Class("yarn", 
+  portable = TRUE,
+  public = list(
+    #' @field path \[`character`\] path to file on disk
```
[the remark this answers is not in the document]
maelle: I am guessing it's a new convention, it slightly rings a bell now that I think about it more :thinking:

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # tinkr dev
 
+* xml and yaml objects are now stored in an R6 class called `yarn`.
```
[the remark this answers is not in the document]
zkamvar: I originally thought of XMD or tink as the class name, but then I realized that there's a perfectly good metaphor just sitting there!

R/class-yarn.R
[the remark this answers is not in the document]
zkamvar: I recall this being done in some rOpenSci packages and I liked the way this identifies the class in the docs (though I'm not committed to keeping this convention, so if you feel strongly that we should not use it, let it be known :)).

R/class-yarn.R
zkamvar: Agreed. maybe a `head()` and `tail()` method that renders to markdown and `cat()`s the output?

R/class-yarn.R
zkamvar: Now that I'm thinking about this, the checking for the path existence should be performed in the `to_md()` function since it's doing all the work here.

R/class-yarn.R
```diff
@@ -0,0 +1,134 @@
...
+    #'   " - This\n - is\n - a new list",
+    #'   where = 2L
+    #' )
+    #' # three headings
+    #' xml2::xml_find_all(ex$body, "md:heading", ex$ns)
+    #' xml2::xml_find_all(ex$body, "md:list", ex$ns)
+    #' tmp <- tempfile()
+    #' ex$write(tmp)
+    #' readLines(tmp, n = 20) 
+    add_md = function(md, where = 0L) {
```
zkamvar, line 121: To keep with the knitting metaphor, we could call this `join_md()` instead, but I think it may be confusing with SQL joins.

zkamvar:
> This is awesome, thank you! I left a few comments.
> 
> Another question that I have is how this should be documented in the README. What approach would we recommend first, R6 or not? In any case at the very top of the README where there's a list of use cases you could add "add an arbitrary new Markdown element like a table".

I think we can recommend the R6 approach first since it's arguably more straightforward given that we are defining the namespace prefix in the object. So the first block would turn into this:

```r
# From Markdown to XML
path <- system.file("extdata", "example1.md", package = "tinkr")
y <- yarn$new(path)

library("magrittr")
# transform level 3 headers into level 1 headers
body <- y$body
ns <- y$ns
body %>%
  xml2::xml_find_all(xpath = './/md:heading[@level="3"]', ns) -> headers3

xml2::xml_set_attr(headers3, "level", 1)

y$body <- body

# Back to Markdown
y$write("newmd.md")
file.edit("newmd")
```

R/class-yarn.R
zkamvar: I've decided to make the path optional for `to_md()` and have it return the md vector, invisibly in 
b97ef02. This way, we can take advantage of `to_md()` for creating the `head()` and `tail()` entries.

zkamvar:
Still to do:

 - README
 - test coverage :see_no_evil:

maelle:
Reg the example I think adding R6 is fine but we might lose some folks if using "->" for assignment :wink:

I wonder whether we should add a short list of things that people might enjoy learning about for maximal enjoyment of tinkr
* XPath
* XSLT (not crucial if they are ok with the default one)
* R6 (although reading the docs of tinkr should help)

R/stylesheet.R
```diff
@@ -0,0 +1,31 @@
+is_stylesheet <- function(stylesheet) {
+  inherits(stylesheet, "xml_document") && 
+    length(xml2::xml_name(stylesheet)) == 1L    &&
+    xml2::xml_name(stylesheet) == "stylesheet"
```
maelle, line 4: so, out-of-scope, but now I wonder how one can actually ensure something is a valid stylesheet.

zkamvar: I would imagine that it would involve an XML schema and `xml2::xml_validate()`, but this is still something that I need to expand my knowledge on.

zkamvar:
> Reg the example I think adding R6 is fine but we might lose some folks if using "->" for assignment :wink:

come to think of it, for this particular example, we don't even need to use any assignment:

```r
y$body %>%
  xml2::xml_find_all(xpath = './/md:heading[@level="3"]', y$ns) %>%
  xml2::xml_set_attr("level", 1)

y$write("newmd.md")
```
 
> I wonder whether we should add a short list of things that people might enjoy learning about for maximal enjoyment of tinkr
> 
>     * XPath
> 
>     * XSLT (not crucial if they are ok with the default one)
> 
>     * R6 (although reading the docs of tinkr should help)

YES! Your original link to the XPath page really helped me learn, so I think these extras would also help. It may also be worthwhile thinking about writing up some documentation about how the XSLT stylesheet in {tinkr} works broadly (e.g. that it uses the commonmark namespace and imports the commonmark stylesheet).

I also realized that it's possible to manipulate a stylesheet with {xml2} so that users can change things like bullet point shape

R/stylesheet.R
maelle: https://github.com/ropensci/tinkr/issues/35

maelle:
I opened a few issues

zkamvar:
This is _very_ strange.... I'm able to get the tests to pass on my machine, even with a clean version of the git repo.

zkamvar:
I think this can be merged that the basic elements of the R6 class have been addressed. I think any other modifications should be new PRs.

PR #34, opened 2020-11-05
print md to screen if path is NULL

maelle:
Or maybe just head and tail... depending on a parameter?

### DIFF
--- a/R/class-yarn.R
+++ b/R/class-yarn.R
@@ -8,7 +8,7 @@
 #' has methods that make it easier to add, analyze, remove, or write elements
 #' of your markdown document.
 #' @export
-yarn <- R6::R6Class("yarn", 
+yarn <- R6::R6Class("yarn",
   portable = TRUE,
   public = list(
     #' @field path \[`character`\] path to file on disk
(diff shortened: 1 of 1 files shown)


zkamvar:
> Or maybe just head and tail... depending on a parameter?

I would prefer to continue to have the `$write()` method return the object, invisibly so that it can be used in a chain: `x$add_md("# title",)$write("file.md")$reset()`

R/class-yarn.R
```diff
@@ -84,15 +84,18 @@ yarn <- R6::R6Class("yarn",
...
       stylesheet_path = system.file("extdata", "xml2md_gfm.xsl", package = "tinkr")){
+
+      output <- to_md(self, path, stylesheet_path)
+
       if (is.null(path)) {
-        stop("Please provide a file path", call. = FALSE)
+        cat(output)
       }
-      to_md(self, path, stylesheet_path)
-      invisible(self)
```
zkamvar, line 93: This part is important for chaining R6 methods (described in the last paragraph of the "Private Members" section: https://r6.r-lib.org/articles/Introduction.html#private-members)

zkamvar (changes requested):
> Or maybe just head and tail... depending on a parameter?

I would prefer to continue to have the `$write()` method return the object, invisibly so that it can be used in a chain: 

```r
x$add_md("# title")$write("file.md")$reset()
```

I think the methods should be semantically concise and avoid the need for several arguments, so having `$write()` both write a file and output markdown seems to be a bit of misnomer. Instead, I think it would be better to have a `$write()` method and a `$head()` and `$tail()` method.

maelle:
I'm a bit lost, what do I need to change?

zkamvar:
Closing this because we came to a consensus that having `head()`, `tail()` and `show()` methods would be better.

PR #39, opened 2021-05-01
add asis node; demonstrate math protection

### LINKED ISSUE #4: tables are not preserved
Thanks for a nice package,

it seems that tables are stripped out such as

```
## tables


--------------------
a    b    c
--- ---  -----------
1    2    3

--------------------
```

is transformed in 

```
## tables

***

a    b    c

***

1    2    3

***
```

using `to_md(to_xml("tinkr.Rmd"), "tinkr2.Rmd")`

zkamvar:
This is definitely still a work in progress!

### DIFF
--- a/R/asis-nodes.R
+++ b/R/asis-nodes.R
@@ -1,0 +1,29 @@
+find_inline_math <- function(body, ns) {
+   i <- ".//md:text[contains(text(), '$') and not(contains(text(), '$$'))]"
+   xml2::xml_find_all(body, i, ns = ns)
+}
+
+find_block_math <- function(body, ns, tag = "md:text[contains(text(), '$$')]") {
+  after  <- "following-sibling::"
+  before <- "preceding-sibling::"
+  after_first_tag <- glue::glue("{after}{tag}")
+  before_last_tag <- glue::glue("{before}md:*[{before}{tag}]")
+  xpath <- glue::glue(".//{after_first_tag}/{before_last_tag}")
+  xml2::xml_find_all(body, xpath, ns = ns)
+}
+
+protect_block_math <- function(body, ns) {
+  bm <- find_block_math(body, ns)
+  xml2::xml_set_attr(bm, "asis", "true")
+}
+
+tick_check <- function(body, ns) {
+  predicate <- "starts-with(text(), '[ ]') or starts-with(text(), '[x]')"
+  cascade <- glue::glue(".//md:item/md:paragraph/md:text[{predicate}]")
+  xml2::xml_find_all(body, cascade, ns = ns)
+}
+
+fix_tickboxes <- function(body, ns) {
+  ticks <- tick_check(body, ns)
+  
+}
--- a/R/class-yarn.R
+++ b/R/class-yarn.R
@@ -163,6 +163,17 @@ yarn <- R6::R6Class("yarn",
       }
       self$body <- copy_xml(b)
       invisible(self)
+    },
+    #' @description Protect math blocks from being escaped
+    #' 
+    #' @examples
+    #' path <- system.file("extdata", "math-example.Rmd", package = "tinkr")
+    #' ex <- tinkr::yarn$new(path)
+    #' ex$tail() # math blocks are escaped :(
+    #' ex$protect_math()$tail() # math blocks are no longer escaped :)
+    protect_math = function() {
+      protect_block_math(self$body, self$ns)
+      invisible(self)
     }
   ),
   private = list(
--- a/inst/extdata/math-example.md
+++ b/inst/extdata/math-example.md
@@ -1,0 +1,40 @@
+---
+title: An example with math elements
+---
+
+
+This example has $\LaTeX$ elements embedded in the
+text. It is intended to demonstrate that m$\alpha\tau$h
+mode can work with tinkr. $y = 
+mx + b$
+
+ - [ ] This is an empty checkbox
+ - [x] This is a checked checkbox
+ - [This is a link](https://ropensci.org)
+ - \[this is an example\]
+
+Here is an example from the mathjax website:
+
+When $a \ne 0$, there are two solutions to \(ax^2 + bx + c = 0\) and they are
+\[x = {-b \pm \sqrt{b^2-4ac} \over 2a}.\]
+
+
+When $a \ne 0$, there are two solutions to $ax^2 + bx + c = 0$ and they are
+
+```latex
+$$
+x = {-b \pm \sqrt{b^2-4ac} \over 2a}
+$$
+```
+
+
+$$
+x = {-b \pm \sqrt{b^2-4ac} \over 2a}
+$$
+
+Below is an example from https://github.com/ropensci/tinkr/issues/38
+
+$$
+Q_{N(norm)}=\frac{C_N +C_{N-1}}2\times 
+\frac{\sum _{i=N-n}^{N}Q_i} {\sum_{j=N-n}^{N}{(\frac{C_j+C_{j-1}}2)}}
+$$
--- a/inst/extdata/xml2md_gfm.xsl
+++ b/inst/extdata/xml2md_gfm.xsl
@@ -16,6 +16,12 @@
 
     <xsl:output method="text" encoding="utf-8"/>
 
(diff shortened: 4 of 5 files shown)


R/add_md.R
```diff
@@ -0,0 +1,53 @@
+#' Add markdown content to an XML object
+#'
+#' @param body an XML object generated via {tinkr}
+#' @param md a string of new markdown to insert
+#' @param where the location in the markdown document to insert the new markdown
```
maelle, line 5: should it be "position"?

R/add_md.R
```diff
@@ -0,0 +1,53 @@
...
+#'
+#' @return a copy of the XML object with the markdown inserted.
+add_md <- function(body, md, where = 0L) {
+  new <- md_to_xml(md)
+  add_node_children(body, new, where)
+}
+
+# Adds children to a specific location and then copies the xml of the full
```
maelle, line 13:
```suggestion
# Add children to a specific location and then copies the xml of the full
```

R/add_md.R
```diff
@@ -0,0 +1,53 @@
...
+# document. 
+add_node_children <- function(body, nodes, where = 0L) {
+  for (child in rev(nodes)) {
```
maelle, line 16: So many clever things in this PR, like this `rev(nodes)` :star_struck:

R/add_md.R
```diff
@@ -0,0 +1,53 @@
...
+  }
+  copy_xml(body)
+}
+
+add_node_siblings <- function(node, nodes, where = "after", remove = TRUE) {
+  for(sib in rev(nodes)) {
+    xml2::xml_add_sibling(node, sib, .where = where)
+  }
+  if (remove) xml2::xml_remove(node)
+}
```
maelle, line 27: why doesn't this function return a copy?

R/asis-nodes.R
```diff
@@ -0,0 +1,199 @@
...
+    # since we split up the nodes, we have to do this node by node
+    for (i in seq(new_nodes)) {
+      add_node_siblings(math[bespoke][[i]], new_nodes[[i]], remove = TRUE)
+    }
+  }
+
+  # protect math that is broken across lines
+  if (length(bmath)) {
+    if (length(bmath[endless]) != length(bmath[headless])) {
+      stop("Uneven math elements")
```
maelle, line 91: codecov tells me there's no test for it, would it make sense to add one?

R/asis-nodes.R
```diff
@@ -0,0 +1,199 @@
...
+  bm <- find_block_math(body, ns)
+  # get all of the internal nodes
+  bm <- xml2::xml_find_all(bm, ".//descendant-or-self::md:*", ns = ns)
+  set_asis(bm) 
+}
+
+# TICK BOXES -------------------------------------------------------------------
+
+tick_check <- function(body, ns) {
+  predicate <- "starts-with(text(), '[ ]') or starts-with(text(), '[x]')"
```
maelle, line 181: out of curiosity, why doesn't the regex contain the "-"/"*"?

maelle (approved):
This is brilliant! :clap: Thank you!

Apart from my small comments/questions, would it make sense to add a paragraph about Math in the README?
Also, when would one not want to use `protect_math()`?

R/asis-nodes.R
zkamvar: Because commonmark has already parsed "-"/"*" as a list item, so we see a hierarchy of list/item/paragraph/text on the xml side.

R/asis-nodes.R
zkamvar: That's a good point. For context, this is here to make sure the number of opening delimiters match the closing delimiters when LaTeX math splits across two lines or contains markdown elements (e.g. `_i` gets converted into `<emph>` if there is a following `_`).

The regex I wrote (`((?=^|[[:space:][:punct:]])[$]?[$][^[:space:]$].+?[^[:space:]$][$][$]?(?=[[:space:][:punct:]]|$))`) has lookaheads to make sure that things like `$2 and $6 are stored in object$thing` don't match and It's difficult to think of a situation where this would be triggered.... unless someone were documenting their BASIC code that parses financial questions without using a code block:

From the first example of: http://qb64.org/wiki/INKEY$

```markdown
Here's a $2 program:

PRINT "Press any keyboard typing key to end SLEEP"
SLEEP
DO: K$ = INKEY$: PRINT K$: LOOP UNTIL K$ = ""
```

NEWS.md
```diff
@@ -3,6 +3,11 @@
 * xml and yaml objects are now stored in an R6 class called `yarn`.
 * testthat edition 3 is now being used with snapshot testing.
 * Tables are now pretty after a full loop `to_xml()` + `to_md()` (@pdaengeli, #9)
+* 2021-05-04: yarn objects remember the `sourcepos` and `encoding` options
```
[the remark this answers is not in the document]
zkamvar: This should be "when using the `$reset()` method"

R/add_md.R
zkamvar:
```suggestion
#' @param where the position in the markdown document to insert the new markdown
```

zkamvar: Yes!

R/add_md.R
zkamvar: Actually, I'm realizing that the companion function, `add_node_children()` should *not* return a copy. 

I do not return a copy here because I use `add_node_siblings()` in a loop so that I can modify multiple nodes in a single document. If I return a copy at this level, then I would have to iterate on that copy. 

For reference, here are all the places in the code where these functions are used. 

```r
R/add_md.R|10| add_node_children(body, new, where)
R/asis-nodes.R|84| add_node_siblings(math[bespoke][[i]], new_nodes[[i]], remove = TRUE)
R/asis-nodes.R|123| add_node_siblings(last_line, nodes, remove = TRUE)
R/asis-nodes.R|195| add_node_siblings(ticks[[i]], new_nodes[[i]], remove = TRUE)

```

R/asis-nodes.R
zkamvar: I've updated this and added an example in 46e9efaad8af1278ae1f6874d20252c73029719

R/add_md.R
zkamvar: I've updated the other function so these both act the same.

zkamvar:
> Apart from my small comments/questions, would it make sense to add a paragraph about Math in the README?

Yes, it would!

> Also, when would one not want to use `protect_math()`?

I can really only think of edge cases (e.g. writing a BASIC tutorial or describing the [Burroughs-Wheeler Transform](https://en.wikipedia.org/wiki/Burrows%E2%80%93Wheeler_transform) with ^ and $ as the delimiters.)

README.md
```diff
@@ -242,10 +238,84 @@ stylesheet](inst/extdata/xml2md_gfm.xsl) and provide its filepath as
...
+    #|       names and code in backtics like so: `INKEY$`. 
+    #| 
+    #| Below are the pairs that were found:
+    #|            start...end
+    #|            -----...---
+    #|  Give you $2 to ... me what INKEY$ means.
+    #|  Give you $2 to ... 2$ verbally.
+    #| We write $2 but ...
+    ```
+
```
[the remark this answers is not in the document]
maelle: it does! I just wonder what folks who do not use the yarn object should do: probably update their code to use the yarn object?

maelle (approved):
Thanks for all the updates!! I only added a small comment.

README.md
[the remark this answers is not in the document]
zkamvar: I think the reason why I am using the yarn object here is because we can use our namespace object in the internal parsing, but I can export that object so that we can use it as default.

```
structure(c(md = "http://commonmark.org/xml/1.0"), class = "xml_namespace")
```

zkamvar:
Excellent! I realized I need to add a couple of more touches to make it complete (e.g. clearing the namespace of the new nodes before copying to [avoid the issue of duplicating namespace](https://community.rstudio.com/t/adding-nodes-in-xml2-how-to-avoid-duplicate-default-namespaces/84870?u=zkamvar)).

Thank you so much for reviewing this!

PR #47, opened 2021-05-25
Support anchored/reference links

### LINKED ISSUE #45: Add option to preserve reference links
I have an idea for {tinkr} that will retain reference links in https://github.com/ropensci/tinkr/commit/0a1ea4b8cafd55d92243fd3cbbac30b940e2d9d8

It's far from resolved, but the idea is that we can add an option in this process that will search for links in the xml document, and then search for `]: <link>` in the body of the source document in `to_xml()` (since we have it in memory at that time). When we find those links, we can add the 'asis' attribute to the links, write a XPath rule, and re-insert the anchor links at the bottom of the document as `asis` text nodes. If we have sourcepos turned on, we could even deduce where these anchors came from and help preserve cases where people define these references per section.

zkamvar:
## Summary

This will allow anchored links in the form of `[text][rel-link]` and
`[rel-link]: <URL> "Title"` to be processed with the folowing caveats:

1. The URL is assumed to have no spaces
2. The anchor will sink to the bottom of the document (until we find a
   good process for re-introducing text nodes into the document).

This will fix #45 
   
### Without anchor link processing

``` r
library("tinkr")
f <- system.file("extdata", "link-test.md", package = "tinkr")
m <- yarn$new(f, anchor_links = FALSE)
m$tail(30) # anchor links unprocesed
#> ---
#> title: this tests links
#> ---
#> 
#> ## These are some links that are valid in basic markdown
#> 
#> This is some text [that contains links](https://example.com/1) which
#> [can be `inline`](https://example.com/2) or [can be spread across multiple lines
#> because the link text is JUST TOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO
#> LONG, y'all](https://example.com/3).
#> 
#> Mainly, we want to see if [link text
#> by reference](https://example.com/4) and if links [can be referenced multiple times](https://example.com/1)
#> 
#> This should also [include non-reference links](https://example.com/5)
#> 
#> If you write \[some link text\]\[link2\], that will appear as [some link text](https://example.com/2 "link with title!")
#> 
#> ## This is some extended markdown content {#extended .callout}
#> 
#> This should also include references that use [standalone](https://example.com/standalone) links and
#> footnotes should not be properly parsed and will be considered 'asis' nodes,
#> at least that's what I *believe*\[^footy\]. Maybe this might not pan out \[^but who
#> knows? footnotes are **WEIRD**, man\].
#> 
#> <!-- links go here! -->
#> 
#> \[^footy\]: this is a footnote that
#> should be preserved
```

### With Anchor link processing

```r
m <- yarn$new(f)
m$tail(30) # anchor links now exist!
#> [can be `inline`][link2] or [can be spread across multiple lines
#> because the link text is JUST TOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO
#> LONG, y'all][link3].
#> 
#> Mainly, we want to see if [link text
#> by reference][link4] and if links [can be referenced multiple times][this fun link1]
#> 
#> This should also [include non-reference links](https://example.com/5)
#> 
#> If you write \[some link text\]\[link2\], that will appear as [some link text][link2]
#> 
#> ## This is some extended markdown content {#extended .callout}
#> 
#> This should also include references that use [standalone][standalone] links and
#> footnotes should not be properly parsed and will be considered 'asis' nodes,
#> at least that's what I *believe*\[^footy\]. Maybe this might not pan out \[^but who
#> knows? footnotes are **WEIRD**, man\].
#> 
#> <!-- links go here! -->
#> 
#> \[^footy\]: this is a footnote that
#> should be preserved
#> 
#> [this fun link1]: https://example.com/1
#> [link2]: https://example.com/2 "link with title!"
#> [link3]: https://example.com/3
#> [link4]: https://example.com/4
#> [standalone]: https://example.com/standalone
```

<sup>Created on 2021-05-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


R/add_md.R
```diff
@@ -14,15 +14,24 @@ add_md <- function(body, md, where = 0L) {
 
 # Add children to a specific location in the full document.
 add_nodes_to_body <- function(body, nodes, where = 0L) {
-  for (child in rev(nodes)) {
-    xml2::xml_add_child(body, child, .where = where)
+  if (inherits(nodes, "xml_node")) {
+    xml2::xml_add_child(body, nodes, .where = where)
+  } else {
+    for (child in rev(nodes)) {
```
maelle, line 20: couldn't we use `purrr::walk()` here?

R/add_md.R
```diff
@@ -14,15 +14,24 @@ add_md <- function(body, md, where = 0L) {
...
 
 # Add siblings to a node
 add_node_siblings <- function(node, nodes, where = "after", remove = TRUE) {
-  for (sib in rev(nodes)) {
-    xml2::xml_add_sibling(node, sib, .where = where)
+  # if there is a single node, then we need only add it
+  if (inherits(nodes, "xml_node")) {
+    xml2::xml_add_sibling(node, nodes, .where = where)
+  } else {
+    for (sib in rev(nodes)) {
```
maelle, line 32: Same comment.

R/asis-nodes.R
```diff
@@ -260,3 +260,38 @@ protect_tickbox <- function(body, ns) {
   }
   copy_xml(body)
 }
+
+# FOOTNOTES --------------------------------------------------------------------
+#nocov start
```
maelle, line 265: why no cov?

R/asis-nodes.R
```diff
@@ -260,3 +260,38 @@ protect_tickbox <- function(body, ns) {
...
+    x = char,
+    perl = TRUE
+  )
+  make_text_nodes(char)
+}
+
+protect_footnotes <- function(body, ns = md_ns()) {
+  body <- copy_xml(body)
+  
+  feet <- footnote_check(body, ns)
```
maelle, line 286: :slightly_smiling_face: :foot:

R/asis-nodes.R
```diff
@@ -260,3 +260,38 @@ protect_tickbox <- function(body, ns) {
   }
   copy_xml(body)
 }
+
+# FOOTNOTES --------------------------------------------------------------------
+#nocov start
+footnote_check <- function(body, ns = md_ns()) {
```
maelle, line 266: why not `check_footnotes()`?

R/resolve-links.R
```diff
@@ -0,0 +1,112 @@
+#' Resolve Reference-Style Links
+#'
+#' Commonmark treats reference-style links as regular links, which can be a
```
maelle, line 3: Maybe add an example or a plain explanation of what reference-style links are? I see the second item in https://github.com/ropensci/tinkr#general-principles-and-solution tries to do that but I wasn't extremely clear at the time.

R/resolve-links.R
```diff
@@ -0,0 +1,112 @@
...
+#' pain when converting large documents. This is an attempt at resolving these
+#' links by reading in the source document, finding the relative links, and 
+#' adding them back with the 'asis' attribute.
+#'
+#' @param body an XML body
+#' @param path path to the source file
+#' @param ns an the NS that resolves the md ns
```
maelle, line 10:
```suggestion
#' @param ns an the namespace that resolves the Markdown namespace
```

R/resolve-links.R
```diff
@@ -0,0 +1,112 @@
...
+#' @param encoding the encoding of the file. Defaults to UTF-8
```
maelle, line 11: are these parameters actually used?

R/resolve-links.R
```diff
@@ -0,0 +1,112 @@
...
+  make_text_nodes(c(
+    "<paragraph>",
+      glue::glue("<link {attrs}>{txt}</link>"),
+    "</paragraph>"
+  ))
+}
+
+# Helpers for parsing anchor links:
+#
+#   [name]: dest 'title'
```
maelle, line 77: very cool to handle link titles too.

R/to_xml.R
```diff
@@ -7,6 +7,9 @@
 #' @param sourcepos passed to [commonmark::markdown_xml()]. If `TRUE`, the
 #'   source position of the file will be included as a "sourcepos" attribute.
 #'   Defaults to `FALSE`.
+#' @param anchor_links if `TRUE`, anchor links (in the style of
```
maelle, line 10: maybe mention what happened to them otherwise (they don't disappear, they're merged into the link(s)).

README.Rmd
```diff
@@ -151,7 +151,8 @@ The (R)md to XML to (R)md loop on which `tinkr` is based is slightly lossy becau
 
  * lists can be created with either "+", "-" or "*". When using `tinkr`, the (R)md after editing will only use "-" for lists. 
  
- * Links built like `[word][smallref]` and bottom `[smallref]: URL` become `[word](URL)`.
+ * Links built like `[word][smallref]` with a bottom anchor `[smallref]: URL`
+  will have the anchor move to the bottom of the document.
```
maelle, line 155:
```suggestion
  will have the anchor moved to the bottom of the document.
```

It seems fitting that anchors sink. :anchor:

maelle (approved):
Brilliant as usual :star2:

R/asis-nodes.R
zkamvar: It was an idea I had and wanted to implement, but I realized it got tricky as trying to keep `[^` and `]` from being escaped might also accidentally escape user-escaped brackets, which look exactly the same to commonmark:

``` r
xml2::read_xml(commonmark::markdown_xml("[ and \\["))
#> {xml_document}
#> <document xmlns="http://commonmark.org/xml/1.0">
#> [1] <paragraph>\n  <text xml:space="preserve">[ and [</text>\n</paragraph>
```

<sup>Created on 2021-05-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

R/resolve-links.R
zkamvar: Good point! I realized there is a good resource for this, actually (and a good source of tests): https://www.markdownguide.org/basic-syntax/#reference-style-links

R/resolve-links.R
zkamvar: Oh! good catch! The doc was written when I was still using the path as the input instead of the lines of text. I will fix this!

README.Rmd
zkamvar: I thought so, too! [The documentation for reference-style links](https://www.markdownguide.org/basic-syntax/#reference-style-links) calls this the "second part" of the link, which I felt was kind of clunky; anchor seemed appropriate. 

And thank you for fixing the grammar! I was going to say that your correction was not right, but then I realized that your edit puts it correctly in future perfect tense and I was trying to do future imperfect.

R/asis-nodes.R
zkamvar: I agree! I had directly copied the `tick_check()` function to make this, and just switched out tick/ticks for footnote/feet :stuck_out_tongue_closed_eyes:

R/add_md.R
zkamvar: I think we could!

R/resolve-links.R
zkamvar: I have vastly expanded the documentation in fb98ee1

zkamvar:
After a few false starts, I have run this against the suite of Software Carpentry lessons without error and I think I should be able to merge this now!

README.Rmd
maelle: Ah you'd know better about grammar than I! I might be wrong!

zkamvar: Au contraire! I speak English natively, which means that you likely know a whole lot more about English grammar than I do! I had to look up what tense I was using.

PR #51, opened 2021-05-26
Clean newly created anchor links

zkamvar:
I ran into a situation where I had this anchor link: `https://www.worldcat.org/search?q=au%3ABenner%2C+Patricia+E.&qt=hot_author`, which XML did not like because  of the `&`. 

This protects the bare ampersands according to https://lachy.id.au/log/2005/10/char-refs by replacing them with `&amp;`


maelle (approved):
Thank you! :pizza: :ice_cream:

Issue #48, opened 2021-05-25
No longer export md_ns() and instead assign the namespace by default?


maelle:

maelle:
What's your take @zkamvar?

zkamvar:
I remember thinking about this earlier and abandoning the idea initially because I didn't really understand how to do it best (you can see my struggles in https://github.com/carpentries/pegboard/blob/main/.notes/namespaces.Rmd)

But now that I think of it, there _is_ a way of doing this that will prevent the need for both `md_ns()` AND `copy_xml()` (maybe), but there are other caveats that we have to deal with. I'm going to write up a short explanation of what I've found and how it impacts {tinkr}, {pegboard}, and any of the other revdeps.

In short, this change will (probably) be a breaking change because the `md:node` prefix will (potentially) no longer work (but that depends on how we handle the situation).

zkamvar:
So, I think the easy solution for us will be the following strategy:

1. strip the namespace immediately on read
2. on write, add the namespace back in as a default namespace

The downsides to this are:

1. all code that used the `md:node` prefix strategy will need to modify the code
2. anyone who exports the raw XML will be missing the crucial namespace, so we might want to provide a wrapper to re-add the default namespace (and no, a prefixed namespace will not work)

maelle:
> all code that used the md:node prefix strategy will need to modify the code

Hopefully that's a breaking change for the best?

> anyone who exports the raw XML will be missing the crucial namespace, so we might want to provide a wrapper to re-add the default namespace (and no, a prefixed namespace will not work)

Why would someone export the raw XML? (genuine question)

maelle:
Of course this is no longer valid after reading the vignette. :grin:

zkamvar:
> Why would someone export the raw XML? (genuine question)

I mean, {tinkr} does some neat parsing with the rmarkdown chunk attributes and anchor links. People might want to export it into XML so they can work with it in a language they understand better (like JS or Python).

PR #52, opened 2021-05-27
Namespace vignette

zkamvar:
This is a work in progress to address #48

### DIFF
--- a/vignettes/namespaces.Rmd
+++ b/vignettes/namespaces.Rmd
@@ -1,0 +1,137 @@
+---
+title: "Hold Namespace For Me"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Hold Namespace For Me}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+## XML namespaces
+
+XML namespaces are a lot like package namespaces in R: they allow you to avoid
+clashes of names [for example, table can represent data or furniture](https://www.w3schools.com/XML/xml_namespaces.asp). 
+
+By default, nodes in XML do not have namespaces unless you give them one, which
+means that when you use XPath search, you can use the node names by default:
+
+```{r}
+d <- xml2::read_xml(glue::glue("<document>
+    <paragraph>
+      <text>hello there</text>
+      <text> ello  here</text>
+    </paragraph>
+  </document>"))
+xml2::xml_ns(d)
+xml2::xml_find_all(d, "//document")
+xml2::xml_find_all(d, "//text[contains(text(), 'hello')]")
+```
+
+However if there is a namespace added to a node, all of its descendants will 
+inherit the namespace, which affects your XPath expressions: 
+
+```{r}
+d <- xml2::read_xml(glue::glue("<document>
+    <paragraph xmlns='http://commonmark.org/xml/1.0'>
+      <text>hello there</text>
+      <text> ello  here</text>
+    </paragraph>
+  </document>"))
+xml2::xml_ns(d)
+xml2::xml_find_all(d, "//document")
+xml2::xml_find_all(d, "//text[contains(text(), 'hello')]") # does not work
+xml2::xml_find_all(d, "//d1:text[contains(text(), 'hello')]") # works; default namespace
+```
+
+When a namespace is specified with `xmlns=<URI>`, {xml2} assigns it a **default
+namespace** prefix, which is `d1`. The {xml2} documentation recommends to
+rename the namespace as soon as you read in a document and use the namespace
+object to semantically prefix your XPath expressions:
+
+```{r}
+ns <- xml2::xml_ns(d)
+ns <- xml2::xml_ns_rename(ns, d1 = "md") 
+ns
+xml2::xml_find_all(d, "//md:text[contains(text(), 'hello')]", ns)
+```
+
+You might be wondering, why don't we prefix the namespace from the start to 
+avoid needing to rename and specify the namespace? Here's an example. Let's take
+our previous example and modify the namespace attribute to have an `md` prefix:
+
+```{r}
+dc <- as.character(d)
+cat(dc <- gsub("xmlns=", "xmlns:md=", dc))
+dc <- xml2::read_xml(dc)
+xml2::xml_ns(dc)
+xml2::xml_find_all(dc, "//document")
+xml2::xml_find_all(dc, "//text[contains(text(), 'hello')]") # works!
+xml2::xml_find_all(dc, "//md:text") # prefix no longer works :(
+```
+
+Now the prefix syntax no longer works, but that's okay, because now we have the
+commonmark as our default namespace, right? Unfortunately, that's not quite the
+case. You might notice that we can access the `document` node AND the `text`
+node without a prefix even though the `text` node is in the commonmark namespace
+and the `document` node is outside of that namespace. It's because neither of
+these nodes actually have a namespace! 
+
+If a namespace is defined in the document with a prefix, *only nodes with that
+prefix are considered to be inside the namespace*. This becomes important when
+we want to pass our XML document through a stylesheet that expects the incoming
+nodes to have a specific namespace, which is exactly how we transform the XML
+representation of markdown back to markdown.
+
+## Commonmark
+
+
+**WIP**
+
+-------------------
+
+We use `commonmark::markdown_xml()` as a starting point to generate valid XML
+text that can be written to a file or read in directly:
+
+```{r cmark}
+cat(cmk <- commonmark::markdown_xml("this is a **test**"))
+xml <- xml2::read_xml(cmk)
+xml
+```
+
+You can see that the first node, `document`, specifies the `xmlns` attribute as 
+`http://commonmark.org/xml/1.0`. This gives a **default namespace**, which means
+that it's used by default across the document, which is useful to us for 
+transforming it back to markdown via our stylesheet:
+
+```{r stysh}
+sty <- xml2::read_xml(tinkr::stylesheet())
+cat(xslt::xml_xslt(xml, sty))
+```
+
+If we look at the stylesheet, we notice the nodes look a bit different---they
+all either have `xsl:` or `md:` prefixes:
+
+```{r}
+sty
+```
+
+If you look at the `stylesheet` node at the top, you'll notice that the 
+namespaces look a bit different. There's no default namespace, but there are two
+namespaces that each have a prefix. 
+
+```{r}
+xml <- xml2::read_xml(gsub("xmlns", "xmlns:md", cmk))  
+xml2::xml_ns(xml)  
+xml2::xml_find_all(xml, ".//md:text") # missing  
+xml2::xml_find_all(xml, ".//text") # okay  
+xml2::xml_ns(xml2::xml_find_all(xml, ".//text"))  
+xml2::xml_add_sibling(xml2::xml_find_first(xml, ".//text"), "text", "there")  
+xml2::xml_find_all(xml, ".//text")  
+```


vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,288 @@
+---
+title: "Hold Namespace For Me"
```
maelle, line 2: It's a nice title but will it be clear enough?

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,288 @@
...
+xml2::xml_find_all(d, "//document")
+xml2::xml_find_all(d, "//text[contains(text(), 'hello')]")
+```
+
+However if there is a namespace added to a node, all of its descendants will 
+inherit the namespace, which affects your XPath expressions: 
+
+```{r}
+d <- xml2::read_xml("<document>
+    <paragraph xmlns='http://commonmark.org/xml/1.0'>
```
maelle, line 52: invalid URL

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,288 @@
...
+      <text> ello  here</text>
+    </paragraph>
+  </document>")
+xml2::xml_ns(d)
+xml2::xml_find_all(d, "//document")
+xml2::xml_find_all(d, "//text[contains(text(), 'hello')]")
+```
+
+However if there is a namespace added to a node, all of its descendants will 
+inherit the namespace, which affects your XPath expressions:
```
maelle, line 48:
```suggestion
inherit the namespace, which affects your XPath expressions.
Below we had the namespace of commonmark to the paragraph node.
```

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,288 @@
...
+```{r}
+d <- xml2::read_xml("<document>
+    <paragraph xmlns='http://commonmark.org/xml/1.0'>
+      <text>hello there</text>
+      <text> ello  here</text>
+    </paragraph>
+  </document>")
+xml2::xml_ns(d)
+xml2::xml_find_all(d, "//document")
+xml2::xml_find_all(d, "//text[contains(text(), 'hello')]") # does not work
```
maelle, line 59: Maybe instead of using code comments the code could be split and commented with text?

````suggestion
Using the same XPath query as before no longer works, our call to `xml2::xml_find_all()` returns nothing.
```{r}
xml2::xml_find_all(d, "//text[contains(text(), 'hello')]") 
````
When a namespace is specified with `xmlns=&lt;URI&gt;`, {xml2} assigns it a **default
amespace** prefix, which is `d1`.
Therefore editing our XPath query like so will work:

```{r}
xml2::xml_find_all(d, "//d1:text[contains(text(), 'hello')]")
```

But is it a good idea to use d1? No, the xml2 docs....

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,288 @@
...
+object to semantically prefix your XPath expressions:
+
+```{r}
+ns <- xml2::xml_ns(d)
+ns <- xml2::xml_ns_rename(ns, d1 = "md") 
+ns
+xml2::xml_find_all(d, "//md:text[contains(text(), 'hello')]", ns)
+```
+
+You might be wondering, why don't we prefix the namespace from the start to
```
maelle, line 75: who's we here? "why isn't it recommended to"

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,288 @@
...
+We use `commonmark::markdown_xml()` as a starting point to generate valid XML:
+
+
+```{r cmark}
+cat(cmk <- commonmark::markdown_xml("this is a **test**"))
+xml <- xml2::read_xml(cmk)
+xml
+```
+
+### It's default of the namespace
```
maelle, line 123: I fear the cool title might be hard to understand for non native speakers.

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,288 @@
...
+```
+
+It works, but let's take a look at our namespaces:
+
+```{r}
+xml2::xml_ns(xml)
+```
+
+Every node we add with an unnamed namespace adds another default and in the end, 
+if we are doing a lot of substitution, we can end up with hundreds of namespaces.
```
maelle, line 229: :scream:

maelle (commented):
Very interesting :nerd_face: :clap: 

My most important comment would be that the vignette lacks an introduction (what's the problem here?) and conclusion.

vignettes/namespaces.Rmd
zkamvar: Good point! What about these:

 1. Why you should care about XML namespaces
 2. Understanding XML namespaces
 3. Fun with XML namespaces
 4. Using XML namespaces in {tinkr}
 3. Namespace: the final frontier
 4. A document by any other namespace would parse just as stylesheet
 
I will stop there because I have other bad puns, but I know this is not what we are aiming for (it is a serious problem for me). I think the title also depends on whether the solution I came up with in #48 works.

vignettes/namespaces.Rmd
zkamvar: This is a product of my own mashed-potato brain not remembering that I'm writing a tutorial :facepalm:

vignettes/namespaces.Rmd
zkamvar: Agreed. I'll change this to "Default Namespaces"

vignettes/namespaces.Rmd
zkamvar: Yeah, this was one of the horrors I stumbled upon when writing {pegboard}.

vignettes/namespaces.Rmd
zkamvar: We could probably obfuscate this by adding a function that creates the URI so that CRAN doesn't pick it up in it's scans.

zkamvar:
> Very interesting nerd_face clap
> 
> My most important comment would be that the vignette lacks an introduction (what's the problem here?) and conclusion.

This is a really good point and really highlights my writing style, it's a bit like building a sandwich inside-out: I start with a salad and add the bread at the end.

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+
+xml2::xml_find_all(xml, "//md:strong", ns) %>%
+  xml2::xml_set_name("code") %>%
+  xml2::xml_set_text("r cat('_test_')")
+
+sty <- xml2::read_xml(tinkr::stylesheet())
+cat(xslt::xml_xslt(xml, sty))
+```
+
+### Perils: adding nodes
```
maelle, line 301: Is "perils" a common word? I understand it because it looks like the French word, but might it be a better idea to use "risks"?

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+
+```{r}
+xml2::xml_set_attr(xml, "xmlns", "http://commonmark.org/xml/1.0")
+xml
+cat(xslt::xml_xslt(xml, sty))
+```
+
+## Harnessing the power of namespaces
+
+When you know that namespaces with prefixes will only respond to nodes with that
```
maelle, line 381: Add a practical use case (in words, not code necessarily) for this?

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+
+While developing {tinkr} we[^1] struggled a lot with understanding namespaces.
+This guide was our attempt at demystifying working with namespaces in {xml2}.
+For the casual user of {tinkr} who is interested in extracting data from 
+markdown documents, this guide is not very useful, but we hope that this 
+guide provies useful for the user who wants to use this for cleaning and
+standardizing their markdown documents. 
+
+[reparse]: https://community.rstudio.com/t/adding-nodes-in-xml2-how-to-avoid-duplicate-default-namespaces/84870/2?u=zkamvar
+[^1]: Well, mostly just Zhian.
```
maelle, line 416: The truth is _you_ took the time to understand them, I just had to read your vignette!

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+xml2::xml_set_attr(xml, "xmlns", "http://commonmark.org/xml/1.0")
+sty <- xml2::read_xml(tinkr::stylesheet())
+cat(xslt::xml_xslt(xml, sty))
+```
+
+## Conclusion
+
+While developing {tinkr} we[^1] struggled a lot with understanding namespaces.
+This guide was our attempt at demystifying working with namespaces in {xml2}.
+For the casual user of {tinkr} who is interested in extracting data from
```
maelle, line 410: You could still write the take-home message for them i.e. what they need to know for normal use?

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+cat(xslt::xml_xslt(xml, sty))
+```
+
+## Conclusion
+
+While developing {tinkr} we[^1] struggled a lot with understanding namespaces.
+This guide was our attempt at demystifying working with namespaces in {xml2}.
+For the casual user of {tinkr} who is interested in extracting data from 
+markdown documents, this guide is not very useful, but we hope that this 
+guide provies useful for the user who wants to use this for cleaning and
```
maelle, line 412:
```suggestion
guide provides useful for the user who wants to use this for cleaning and
```

maelle (commented):
Awesome work, mostly nitpicky comments. :smiley_cat:

vignettes/namespaces.Rmd
zkamvar:
```suggestion
title: "Understanding XML Namespaces"
```

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+
+This document was written to address common confusions about XML namespaces and
+their implications in constructing XPath queries, adding new XML nodes, and 
+converting XML to markdown. This guide is written for the user who is 
+comfortable with XPath queries and wants to understand more about how to handle
+and manpiulate their XML representation of markdown.
+
+### Motivation 
+
+The underlying motivation for {tinkr} was to wrap the process of converting
```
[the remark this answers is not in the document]
zkamvar: I think the curly braces for package names helps avoid potential confusion between talking about XML concepts and R concepts.

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+```
+
+We use the `xslt` package to to the conversion from XML back to markdown.
+
+```{r}
+xslt_style <- tinkr::stylesheet() %>% xml2::read_xml()
+cat(xslt::xml_xslt(xml, xslt_style))
+```
+
+One of the downsides of this conversion is that commonmark provides a default
```
zkamvar, line 53:
```suggestion
One of the downsides of this conversion is that {commonmark} provides a default
```

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+```
+
+Using the same XPath query as before no longer works, our call to
+`xml2::xml_find_all()` returns nothing.
+
+```{r}
+xml2::xml_find_all(d, "//text[contains(text(), 'hello')]") # does not work
+```
+
+When a namespace is specified with `xmlns=&lt;URI&gt;`, {xml2} assigns it a
```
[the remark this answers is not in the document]
zkamvar: Oh, that's the way it parsed in your suggestion. I'll change it back. 
```suggestion
When a namespace is specified with `xmlns=<URI>`, {xml2} assigns it a
```

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+
+You might be wondering, why isn't it recommended to prefix the namespace from
+the start to avoid needing to rename and specify the namespace? The reason is
+because the *prefixed namespaces only apply to nodes with that prefix*. Here's
+an example. Let's take our previous example and modify the namespace attribute
+to have an `md` prefix:
+
+```{r}
+dc <- as.character(d)
+cat(dc <- gsub("xmlns=", "xmlns:md=", dc))
```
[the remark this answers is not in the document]
zkamvar: Good point. It is valid, but weird R to do this.

vignettes/namespaces.Rmd
```diff
@@ -0,0 +1,416 @@
...
+```{r}
+xml2::xml_find_all(dc, "//md:text") # one node
+xml2::xml_find_all(dc, "//text")    # two nodes
+```
+
+If a namespace is defined in the document with a prefix, *only nodes with that
+prefix are considered to be inside the namespace*. This becomes important when
+we want to pass our XML document through a stylesheet that expects the incoming
+nodes to have a specific namespace, which is exactly how we transform the XML
+representation of markdown back to markdown.
```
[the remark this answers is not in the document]
zkamvar: :monocle_face:

vignettes/namespaces.Rmd
zkamvar: I think peril gives a better sense of "something that can not be avoided on this path" as opposed to risk, which has a random component.

vignettes/namespaces.Rmd
zkamvar: Good question! I use the alternate namespace in {pegboard} to help me identify and label pandoc fenced-div sections by adding pairs of equivalently labeled tags that are not part of the markdown document so I can easily parse the content with `find_between()`

Otherwise, I can see the masking pattern useful if you wanted to create several versions of the same prose in a single document (e.g. if you were creating a quiz that you wanted randomized per student).

vignettes/namespaces.Rmd
zkamvar:
```suggestion
guide proves useful for the user who wants to use this for cleaning and
```

PR #54, opened 2021-09-15
Fix 53

### LINKED ISSUE #53: BUG: numeric chunk options are accidentally quoted
If you specify a chunk option with a numeric input such as `fig.width=4.2`, it becomes quoted on output, which causes knitr to error:

``` r
library(tinkr)
f <- tempfile()
writeLines("```{r ggplot, fig.width=4.2}\n# code here\n```", f)
yarn$new(f)$show()
#> ```{r ggplot, fig.width="4.2"}
#> # code here
#> ```
```

<sup>Created on 2021-09-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Interestingly enough, this is a problem on the md -> XML side, not on the XML -> md side. 

For example, here is the boomer result of `transform_params()`:

````r
> library(boomer)
> devtools::load_all()                                                                                  
ℹ Loading tinkr

> rig_in_namespace(transform_params)                                                                    

> f <- tempfile() 
  writeLines("```{r ggplot, fig.width=8.8}\n# code here\n```", f) 
  yarn$new(f)$show()                                                                                    
👇 transform_params
💣 try(eval(parse(text = paste("alist(", quote_label(params), ")"))),... 
· 💣 eval(parse(text = paste("alist(", quote_label(params), ")"))) 
· · 💣 parse(text = paste("alist(", quote_label(params), ")")) 
· · · 💣 paste("alist(", quote_label(params), ")") 
· · · · params :
· · · · [1] "r ggplot, fig.width=8.8"
· · · · 💣 💥 quote_label(params) 
· · · · [1] "'r ggplot', fig.width=8.8"
· · · · 
· · · 💥 paste("alist(", quote_label(params), ")") 
· · · [1] "alist( 'r ggplot', fig.width=8.8 )"
· · · 
· · 💥 parse(text = paste("alist(", quote_label(params), ")")) 
· · expression(alist( 'r ggplot', fig.width=8.8 ))
· · 
· 💥 eval(parse(text = paste("alist(", quote_label(params), ")"))) 
· [[1]]
· [1] "r ggplot"
· 
· $fig.width
· [1] 8.8
· 
· 
💥 try(eval(parse(text = paste("alist(", quote_label(params), ")"))),
     silent = TRUE
   ) 
[[1]]
[1] "r ggplot"

$fig.width
[1] 8.8


💣 if (inherits(params_string, "try-error")) {... 
· 💣 💥 inherits(params_string, "try-error") 
· [1] FALSE
· 
💥 if (inherits(params_string, "try-error")) {
     params <- stringr::str_replace(params, " ", ", ")
     params_string <- eval(parse(text = paste(
       "alist(", quote_label(params),
       ")"
     )))
   } 
NULL

💣 parse_label(params_string[[1]]) 
· 💣 💥 params_string[[1]] 
· [1] "r ggplot"
· 
💥 parse_label(params_string[[1]]) 
$language
[1] "r"

$name
[1] "ggplot"


💣 unlist(c(label, params_string[-1])) 
· 💣 c(label, params_string[-1]) 
· · 💣 params_string[-1] 
· · · 💣 💥 -1 
· · · [1] -1
· · · 
· · 💥 params_string[-1] 
· · $fig.width
· · [1] 8.8
· · 
· · 
· 💥 c(label, params_string[-1]) 
· $language
· [1] "r"
· 
· $name
· [1] "ggplot"
· 
· $fig.width
· [1] 8.8
· 
· 
💥 unlist(c(label, params_string[-1])) 
 language      name fig.width 
      "r"  "ggplot"     "8.8" 

· 💣 result[purrr::map_lgl(result, is.character) & !names(result) %in%... 
· · 💣 purrr::map_lgl(result, is.character) & !names(result) %in% c(... 
· · · · 💣 .f(.x[[i]], ...) 
· · · · 💥 .f(.x[[i]], ...) 
· · · · [1] TRUE
· · · · 
· · · · 💣 .f(.x[[i]], ...) 
· · · · 💥 .f(.x[[i]], ...) 
· · · · [1] TRUE
· · · · 
· · · · 💣 .f(.x[[i]], ...) 
· · · · 💥 .f(.x[[i]], ...) 
· · · · [1] TRUE
· · · · 
· · · 💣 💥 purrr::map_lgl(result, is.character) 
· · ·  language      name fig.width 
· · ·      TRUE      TRUE      TRUE 
· · · 
· · · 💣 !names(result) %in% c("language", "name") 
· · · · 💣 names(result) %in% c("language", "name") 
· · · · · 💣 💥 names(result) 
· · · · · [1] "language"  "name"      "fig.width"
· · · · · 
· · · · · 💣 💥 c("language", "name") 
· · · · · [1] "language" "name"    
· · · · · 
· · · · 💥 names(result) %in% c("language", "name") 
· · · · [1]  TRUE  TRUE FALSE
· · · · 
· · · 💥 !names(result) %in% c("language", "name") 
· · · [1] FALSE FALSE  TRUE
· · · 
· · 💥 purrr::map_lgl(result, is.character) & !names(result) %in% c(
· ·      "language",
· ·      "name"
· ·    ) 
· ·  language      name fig.width 
· ·     FALSE     FALSE      TRUE 
· · 
· 💥 result[purrr::map_lgl(result, is.character) & !names(result) %in%
·      c("language", "name")] 
· fig.width 
·     "8.8" 
· 
💣 💥 glue::glue("\"{result[purrr::map_lgl(result, is.character)&\n           !names(result) %in% c(\"language\", \"name\")]}\"") 
"8.8"

💣 purrr::map_lgl(result, is.character) & !names(result) %in% c(... 
· · 💣 .f(.x[[i]], ...) 
· · 💥 .f(.x[[i]], ...) 
· · [1] TRUE
· · 
· · 💣 .f(.x[[i]], ...) 
· · 💥 .f(.x[[i]], ...) 
· · [1] TRUE
· · 
· · 💣 .f(.x[[i]], ...) 
· · 💥 .f(.x[[i]], ...) 
· · [1] TRUE
· · 
· 💣 💥 purrr::map_lgl(result, is.character) 
·  language      name fig.width 
·      TRUE      TRUE      TRUE 
· 
· 💣 !names(result) %in% c("language", "name") 
· · 💣 names(result) %in% c("language", "name") 
· · · 💣 💥 names(result) 
· · · [1] "language"  "name"      "fig.width"
· · · 
· · · 💣 💥 c("language", "name") 
· · · [1] "language" "name"    
· · · 
· · 💥 names(result) %in% c("language", "name") 
· · [1]  TRUE  TRUE FALSE
· · 
· 💥 !names(result) %in% c("language", "name") 
· [1] FALSE FALSE  TRUE
· 
💥 purrr::map_lgl(result, is.character) & !names(result) %in% c(
     "language",
     "name"
   ) 
 language      name fig.width 
    FALSE     FALSE      TRUE 

💣 💥 c(
        "echo", "eval", "collapse", "warning", "error", "message",
        "split", "include", "strip.white", "prompt", "highlight",
        "cache", "autodep", "external", "sanitize", "purl"
      ) 
 [1] "echo"        "eval"        "collapse"    "warning"     "error"       "message"     "split"      
 [8] "include"     "strip.white" "prompt"      "highlight"   "cache"       "autodep"     "external"   
[15] "sanitize"    "purl"       

💣 if (any(names(result) %in% logical_options)) {... 
· 💣 any(names(result) %in% logical_options) 
· · 💣 names(result) %in% logical_options 
· · · 💣 💥 names(result) 
· · · [1] "language"  "name"      "fig.width"
· · · 
· · 💥 names(result) %in% logical_options 
· · [1] FALSE FALSE FALSE
· · 
· 💥 any(names(result) %in% logical_options) 
· [1] FALSE
· 
💥 if (any(names(result) %in% logical_options)) {
     result[names(result) %in% logical_options & is.character(result)] <- stringr::str_remove_all(result[names(result) %in%
       logical_options & is.character(result)], "\\\"")
   } 
NULL

👆 transform_params
```{r ggplot, fig.width="8.8"}
# code here
```

````

### PR DESCRIPTION
**Pull request by zkamvar**
I've modified `transform_params()` in the following ways:

1. assigned the subsetting statements to descriptive names
2. disallowed logical params from being quoted in the first place
3. protected fig.width and fig.height from being quoted
4. replaced the {glue} statement with the equivalent `shQuote(var, type = "cmd")`

reprex:

``` r
library(tinkr)
f <- tempfile()
writeLines("```{r ggplot, fig.width=4.2}\n# code here\n```", f)
yarn$new(f)$show()
#> ```{r ggplot, fig.width=4.2}
#> # code here
#> ```
```

<sup>Created on 2021-09-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This will fix #53

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -57,33 +57,38 @@ yaml_load = function(x) yaml::yaml.load(
 
 # from knitr via namer
 # https://github.com/lockedata/namer/blob/2d88c5cb200724f775631946fc8e08903ff110de/R/utils.R#L3
-transform_params <- function(params){
-  params_string <- try(eval(parse(text = paste('alist(', quote_label(params), ')'))),
-                       silent = TRUE)
+transform_params <- function(params) {
 
-  if (inherits(params_string, "try-error")){
+  # Step 1: parse the parameters and their labels into a list
+  params_list <- try(parse_params(params), silent = TRUE)
+
+  if (inherits(params_list, "try-error")){
     params <- stringr::str_replace(params, " ", ", ")
-    params_string <- eval(parse(text = paste('alist(', quote_label(params), ')')))
+    params_list <- parse_params(params)
   }
 
-  label <- parse_label(params_string[[1]])
+  label <- parse_label(params_list[[1]])
+  result <- c(label, params_list[-1])
 
-  result <- unlist(c(label, params_string[-1]))
-
-  logical_options  <- c("echo", "eval", "collapse", "warning",
-                       "error", "message", "split", "include",
-                       "strip.white", "prompt", "highlight", "cache",
-                       "autodep", "external", "sanitize", "purl")
-  unquoted_options <- c("language", "name", "fig.width", "fig.height")
-
+  # Step 2: find the parameters that are characters because we need to add
+  # quotes around them (as all parameters are coerced as characters)
   are_characters <- purrr::map_lgl(result, is.character)
-  not_forbidden  <- !names(result) %in% c(unquoted_options, logical_options)
+
+  # Step 3: flatten all params into a character vector
+  result <- unlist(result)
+
+  # Step 4: add quotes around the params that are characters
+  not_forbidden <- !names(result) %in% c("language", "name")
   needs_quoting  <- are_characters & not_forbidden
   result[needs_quoting] <- shQuote(result[needs_quoting], type = "cmd")
 
   result
+
 }
 
+parse_params <- function(params) {
+  eval(parse(text = paste('alist(', quote_label(params), ')')))
+}
 
 parse_label <- function(label){
   label %>%
--- a/R/utils.R
+++ b/R/utils.R
@@ -62,7 +62,7 @@ transform_params <- function(params) {
   # Step 1: parse the parameters and their labels into a list
   params_list <- try(parse_params(params), silent = TRUE)
 
-  if (inherits(params_list, "try-error")){
+  if (inherits(params_list, "try-error")) {
     params <- stringr::str_replace(params, " ", ", ")
     params_list <- parse_params(params)
   }
@@ -79,7 +79,7 @@ transform_params <- function(params) {
 
   # Step 4: add quotes around the params that are characters
   not_forbidden <- !names(result) %in% c("language", "name")
-  needs_quoting  <- are_characters & not_forbidden
+  needs_quoting <- are_characters & not_forbidden
   result[needs_quoting] <- shQuote(result[needs_quoting], type = "cmd")
 
   result
--- a/R/utils.R
+++ b/R/utils.R
@@ -61,7 +61,7 @@ transform_params <- function(params){
   params_string <- try(eval(parse(text = paste('alist(', quote_label(params), ')'))),
                        silent = TRUE)
 
-  if(inherits(params_string, "try-error")){
+  if (inherits(params_string, "try-error")){
     params <- stringr::str_replace(params, " ", ", ")
     params_string <- eval(parse(text = paste('alist(', quote_label(params), ')')))
   }
@@ -69,25 +69,19 @@ transform_params <- function(params){
   label <- parse_label(params_string[[1]])
 
   result <- unlist(c(label, params_string[-1]))
-  result[purrr::map_lgl(result, is.character)&
-           !names(result) %in% c("language", "name")] <- glue::glue('"{result[purrr::map_lgl(result, is.character)&
-           !names(result) %in% c("language", "name")]}"')
 
-  logical_options <- c("echo", "eval", "collapse", "warning",
+  logical_options  <- c("echo", "eval", "collapse", "warning",
                        "error", "message", "split", "include",
                        "strip.white", "prompt", "highlight", "cache",
                        "autodep", "external", "sanitize", "purl")
-  if(any(names(result) %in% logical_options)){
+  unquoted_options <- c("language", "name", "fig.width", "fig.height")
 
-    result[names(result) %in% logical_options&
-             is.character(result)] <- stringr::str_remove_all(result[names(result) %in% logical_options&
-                                                                       is.character(result)],
-                                                                          "\\\"")
+  are_characters <- purrr::map_lgl(result, is.character)
+  not_forbidden  <- !names(result) %in% c(unquoted_options, logical_options)
+  needs_quoting  <- are_characters & not_forbidden
+  result[needs_quoting] <- shQuote(result[needs_quoting], type = "cmd")
 
-
-  }
-
-   result
+  result
 }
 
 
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -61,33 +61,27 @@ transform_params <- function(params){
   params_string <- try(eval(parse(text = paste('alist(', quote_label(params), ')'))),
                        silent = TRUE)
 
-  if(inherits(params_string, "try-error")){
+  if (inherits(params_string, "try-error")){
```
maelle, line 64:
```suggestion
  if (inherits(params_string, "try-error")) {
```

Even more whitespace :grin:

R/utils.R
```diff
@@ -61,33 +61,27 @@ transform_params <- function(params){
...
 
   label <- parse_label(params_string[[1]])
 
   result <- unlist(c(label, params_string[-1]))
-  result[purrr::map_lgl(result, is.character)&
-           !names(result) %in% c("language", "name")] <- glue::glue('"{result[purrr::map_lgl(result, is.character)&
-           !names(result) %in% c("language", "name")]}"')
 
-  logical_options <- c("echo", "eval", "collapse", "warning",
+  logical_options  <- c("echo", "eval", "collapse", "warning",
```
maelle, line 73:
```suggestion
  logical_options <- c("echo", "eval", "collapse", "warning",
```

maelle (approved):
Thank you for catching and fixing this!

I only wonder whether we should add two options for letting users grow the list of unquoted and logical parameters (in case they are dealing with documents where they defined and use parameters)? Or [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)*?

*: phrase I learnt last week. :sweat_smile:

zkamvar:
> I only wonder whether we should add two options for letting users grow the list of unquoted and logical parameters (in case they are dealing with documents where they defined and use parameters)? Or YAGNI*?

I'm actually thinking that we could simplify this a bit more. We only need to check for character arguments so that we can add an extra layer of quotes. Instead of explicitly creating a list of params to _not_ quote, we could map over the initial list (before unlisting) and find the characters that need quoting. I'll add another commit for this.

R/utils.R
zkamvar: good catch!

zkamvar:
Because I changed a part of the code that was not previously covered by tests and now that I changed it, codecov is complaining because this line is not being covered :weary: It would be good to get this line covered. Do you know what situation we would run into this? it seems to be a missing comma issue, but I'm unsure of how to trigger it.

maelle:
sorry which line is that exactly? I think I'm not looking at the right thing.

zkamvar:
> sorry which line is that exactly? I think I'm not looking at the right thing.

It's line 67 which is only run if the first pass through parsing params fails: https://github.com/ropensci/tinkr/pull/54/files#diff-60bef5af43a1bd541299ce9bceeb226afb23deaa5e637b145c85e05ffd552b0cR67

maelle:
mmmh, fine to ignore?

zkamvar:
> mmmh, fine to ignore?

I think so. In the future, it would be good to revisit this to figure out what situations would cause this block to be called.

PR #73, opened 2022-08-08
cran release 010

zkamvar:
This PR will be merged once tinkr is released on CRAN


zkamvar:
Got some feedback:

```
Please do not start the description with "This package", package name,
title or similar.

"Using foo:::f instead of foo::f allows access to unexported objects.
This is generally not recommended, as the semantics of unexported
objects may be changed by the package author in routine maintenance."
Please omit one colon.

You have examples for unexported functions. -> man/yarn.Rd
Please either omit these examples or export these functions.

Please choose a more meaningful vignette title and not "Untitled".

Please ensure that your functions do not write by default or in your
examples/vignettes/tests in the user's home filespace (including the
package directory and getwd()). This is not allowed by CRAN policies.
Please omit any default path in writing functions.
In your examples/vignettes/tests you can write to tempdir().

Please fix and resubmit.
```

A lot of this is not relevant to {tinkr} (e.g description and vignette) and I have no clue why they got inserted, but some of this is (using non-exported functions in help documentation to document those functions), but it oddly points to the wrong documentation file. 

I think the tests point is legit, though, so I'll have a looksee.

zkamvar:
Here is an example of a test file that does not follow the normal tempfile conventions: https://github.com/ropensci/tinkr/blob/d9330012ad93f44517eb4f93ee62c03f5de94a3c/tests/testthat/test-to_md.R#L25-L27

zkamvar:
Response from CRAN: 

```
Thanks,

 > Please do not start the
    description with "This package", package name, title
    or similar.

Your description starts with:
     "Casts '(R)Markdown' files to 'XML' and back to..."
Which is essentially your title.

 >Please choose a more meaningful vignette title and
   not "Untitled".

While you don't have an untitled vignette, there are a lot of .Rmd files
in your /inst and /test directories which seem to be unnecessary. Please
check if you really need them all or if some of them can be omitted.

Please fix and resubmit.
```

zkamvar:
Note: I'm going to merge this because, while it is _not_ complete, I got the following comment from Uwe:

```
Thanks, we see:

   Found the following (possibly) invalid URLs:
     URL:
https://github.com/ropensci/tinkr/blob/main/inst/stylesheets/xml2md_gfm.xsl
       From: inst/doc/tinkr.html
             README.md
       Status: 404
       Message: Not Found

Please fix and resubmit.
```


And of course this URL will be valid once it is merged into main. So I'm going
to decrement the version number, merge, change back to this branch, and increment
the version number again and submit.

PR #76, opened 2022-09-15
feat: protect_curly()

### LINKED ISSUE #75: Protect text inside curly braces?
In text such as

```
# Preface {.unnumbered}
```

it might be nice if the curly braces ended up in a special tag.

But we'd be limited by commonmark @zkamvar? :thinking:

maelle:
Fix #75

I'll be happy to learn while working on this PR :smile_cat:


R/asis-nodes.R
```diff
@@ -320,3 +320,57 @@ protect_footnotes <- function(body, ns = md_ns()) {
   copy_xml(body)
 }
 #nocov end
+# CURLY ------------------
+find_curly <- function(body, ns) {
+   i <- ".//md:text[not(@asis) and contains(text(), '{')]"
+   xml2::xml_find_all(body, i, ns = ns)
+}
+
+fix_curly <- function(curly) {
```
maelle, line 329: clearly my regex level is not on par with this file :sweat_smile:

R/asis-nodes.R
```diff
@@ -320,3 +320,57 @@ protect_footnotes <- function(body, ns = md_ns()) {
...
+  char <- as.character(curly)
+  curlies <- regmatches(char, regexpr("\\{.*?\\}", char))
+  for (curl in curlies) {
+    char <- sub(
+      curl,
+      sprintf("</text><text asis='true' curly='true'>%s</text><text>", curl),
```
maelle, line 335: I also added the curly attribute because it might be useful to differentiate the asis node -- is that a bad idea?

tests/testthat/test-asis-nodes.R
```diff
@@ -78,3 +78,8 @@ test_that("documents with no math do no harm", {
   expect_equal(as.character(x), x1)
 })
 
+test_that("mal-formed inline math throws an informative error", {
```
zkamvar, line 81:
```suggestion
test_that("things inside curly braces get protected", {
```

R/asis-nodes.R
zkamvar: I call BS on this comment! I have never known how to work with the regmatches and gregexpr functions and I commend your ability to do this!

R/asis-nodes.R
zkamvar: Nope! I think it's in line with what we want to do. In fact, I don't think we need the `asis` node here because we do not necessarily need to protect things inside curlies, unless there is something I am missing.

inst/extdata/basic-curly.md
```diff
@@ -0,0 +1,9 @@
+---
+title: basic curly
+---
+
+# preface {#pre-face .unnumbered}
+
+hello
+
+I like {xml2} but of course {tinkr} is even cooler!
```
zkamvar, line 9: One of the difficulties with this approach is that it will also capture and ignore image alt text, which _should_ be sent to the API, whereas the other attributes should be left untouched. 

```suggestion
I like {xml2} but of course {tinkr} is even cooler!

Images that use pandoc style will have curlies with content that should be translated and should be protected.

![a pretty kitten](https://placekitten.com/200/300){#kitteh alt='a picture of a kitten'}

```

R/asis-nodes.R
```diff
@@ -320,3 +320,57 @@ protect_footnotes <- function(body, ns = md_ns()) {
...
+  }
+  make_text_nodes(char)
+}
+
+#' Protect curly elements for further processing
+#'
+#' @inheritParams protect_math
+#' @return a copy of the modified XML object
+#' @details Commonmark will render text such as `{.unnumbered}`
+#' (Quarto option) or
```
zkamvar, line 348:
```suggestion
#' (Pandoc/Quarto option) or
```

zkamvar (commented):
This looks good! It works as it says on the tin and will catch most of the cases. I'm also thinking that this could be generalised into a function that can take a generic opening pattern and closing pattern to give a specific label so that people can extend it for their own purposes. 


There are a couple of cases where the solution at this stage would produce undesired results:

1. If attributes are split across lines, it will only find the first line:
   ```
   [a span with attributes]{.span-with-attributes 
   style='color: red;'}
   ```
   The solution is to search one of the following siblings of the nodes for a closing brace [as I've done in pegboard](https://github.com/carpentries/pegboard/blob/a8db02ab037f2ffeab6e13cc3b662ea8c42822ad/R/get_images.R#L66-L97).
2. Images with alt text will be captured
   ```
   ![caption](img.png){alt='alternative text for the image'}
   ```
   I'm not sure what the solution to this might be, but I suspect we would need to extract this alt text and put it in an attribute. 
   
I'd be happy to try my hand at this, but I'm also happy to merge this as an "in progress" feature before I make an update to CRAN in a couple of months.

NAMESPACE
```diff
@@ -2,6 +2,7 @@
 
 export(find_between)
 export(md_ns)
+export(protect_curly)
```
maelle, line 5: I opened an issue regarding the generalization #77

R/attr-nodes.R
```diff
@@ -0,0 +1,83 @@
+# CURLY ------------------
```
maelle, line 1: I wasn't sure how to name this file.

R/attr-nodes.R
```diff
@@ -0,0 +1,83 @@
...
+  no_closing <- !grepl("[}]", attr_texts)
+  if (any(no_closing)) {
+    close_xpath <- "self::*/following-sibling::md:text[contains(text(), '}')]"
+    for (not_closed in curlies[no_closing]) {
+      closing <- xml2::xml_find_all(
+        not_closed,
+        glue::glue("./{close_xpath}"),
+        ns
+      )
+      xml2::xml_text(not_closed) <- paste(
```
maelle, line 16: should I add a test for the md -> xml + protect_curly() -> md round? If so is there an example?

R/asis-nodes.R
maelle: I can't even remember how I got it to work tbh :joy_cat:

R/attr-nodes.R
zkamvar: Yes, we should absolutely test this to make sure nothing on the xslt side suddenly changes. I think the easiest way to do this would be to use a file in a temporary directory and check the snapshot of the file:

https://github.com/ropensci/tinkr/blob/a577e9b4f3b94bcf09ead5c77ed49373a31db5d6/tests/testthat/test-class-yarn.R#L53-L63

Also, this is a clever solution 🤩! I had not thought about smashing together text into a text node separated by a newline like this! I always made them separate text nodes, but this solution solves so many problems.

inst/extdata/basic-curly.md
```diff
@@ -0,0 +1,19 @@
...
+
+Images that use pandoc style will have curlies with content that should be translated and should be protected.
+
+![a pretty kitten](https://placekitten.com/200/300){#kitteh alt='a picture of a kitten'}
+
+![a pretty puppy](https://placedog.net/200/300){#dog alt="a picture
```
zkamvar, line 15: 🐶

tests/testthat/_snaps/attr-nodes.md
```diff
@@ -0,0 +1,53 @@
...
+        <paragraph>
+          <image destination="https://placedog.net/200/300" title="">
+            <text xml:space="preserve">a pretty puppy</text>
+          </image>
+          <text xml:space="preserve"/>
+          <text curly="true" alt="a picture of a dog">{#dog alt="a picture
+      of a dog"}</text>
+          <text/>
+          <softbreak/>
+        </paragraph>
```
zkamvar, line 44: I think this softbreak was originally between the text nodes, but I think it's benign

R/attr-nodes.R
zkamvar: Considering how inconsistently I was at naming `asis-nodes` and other files, this fits in with the general aesthetic 😅

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# tinkr (development version)
+
+* New `protect_curly()` function.
```
zkamvar, line 3:
```suggestion
* New `protect_curly()` function will add a `curly='true'` attribute to text wrapped in curly braces ('{', '}') to allow parsing of the XML for sending to external APIs. This function will also parse alt text and place it in an attribute.
* New `$protect_curly()` method implements `protect_curly()` on yarn objects
```

zkamvar (commented):
YES! Thank you for the quick changes. The only thing left is the tests!

inst/extdata/basic-curly.md
```diff
@@ -0,0 +1,19 @@
+---
+title: basic curly
```
maelle, line 2: why is YAML added when there was originally no YAML? is this expected?

tests/testthat/_snaps/attr-nodes/yarn.md
```diff
@@ -0,0 +1,23 @@
...
+
+hello
+
+I like {xml2} but of course {tinkr} is even cooler!
+
+Images that use pandoc style will have curlies with content that should be translated and should be protected.
+
+![a pretty kitten](https://placekitten.com/200/300){#kitteh alt='a picture of a kitten'}
+
+![a pretty puppy](https://placedog.net/200/300){#dog alt="a picture
```
maelle, line 15: missing space at the end.

tests/testthat/_snaps/attr-nodes/yarn.md
```diff
@@ -0,0 +1,23 @@
...
+of a dog"}
+
+
+\[a span with attributes\]{.span-with-attributes
```
maelle, line 19: is this escaping expected?

R/attr-nodes.R
maelle: I was not sure whether it was a clever or lazy solution to be honest :joy:

tests/testthat/_snaps/attr-nodes/yarn.Rmd
```diff
@@ -0,0 +1,30 @@
...
+
+![a pretty kitten](https://placekitten.com/200/300){#kitteh alt='a picture of a kitten'}
+
+![a pretty puppy](https://placedog.net/200/300){#dog alt="a picture
+of a dog"}
+
+
+\[a span with attributes\]{.span-with-attributes
+style='color: red;'}
+
```
maelle, line 28: why so many new lines

zkamvar: I believe it's because of the `<softbreak>`: https://github.com/ropensci/tinkr/pull/76/files/3e133e5f6e6b35befef6d8fad20df6d87f8ad549?w=0#r973214227. I guess I was wrong in calling it benign -_-

tests/testthat/_snaps/attr-nodes/yarn.md
zkamvar: Unfortunately, yes: https://github.com/r-lib/commonmark/issues/20 

I'm going to try to implement this today. It should be a matter of backtracking and finding the square braces that were not escaped in the original document and then tagging them with 'asis' tags.

zkamvar (approved):
I'm happy to merge this for now given the current limitations and then work on fixing the unescaped square braces.

maelle:
So I'll let you finish this PR? Alternatively I can work on it again later this week. Thank you!!

zkamvar:
I've opened #78 to fix the unescaped braces issue, and I think the extra space is a minor issue that we can fix before the CRAN release, so I'm going to merge this.

PR #78, opened 2022-09-20
protect escaped brackets

zkamvar:
This provides a patch to avoid escaping unescaped brackets when doing a roundtrip.

``` r
tmp <- tempfile()
writeLines("\\[escaped\\] and [unescaped] and [links](https://example.com) will be respected", tmp)
```

No sourcepos, no protection for unescaped braces

``` r
tinkr::yarn$new(tmp)$show()
#> \[escaped\] and \[unescaped\] and [links](https://example.com) will be respected
```

Add sorucepos, then protection happens

``` r
tinkr::yarn$new(tmp, sourcepos = TRUE)$show()
#> \[escaped\] and [unescaped] and [links](https://example.com) will be respected
```

<sup>Created on 2022-09-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


NEWS.md
```diff
@@ -1,5 +1,9 @@
 # tinkr (development version)
 
+* unescaped square braces from the source code can now make the roundtrip
+  without extra markup. This is implemented in a new default parameter for
+  `to_xml()` called `unescaped = TRUE`, which indicates that you would like
```
maelle, line 5: should it be called something more specific? `protect_unescaped_square_braces` is long but clear :thinking:

NEWS.md
```diff
@@ -1,5 +1,9 @@
...
+  unescaped square braces to remain unescaped.
```
maelle, line 6:
```suggestion
  unescaped square braces to remain unescaped (#78). 
```

R/asis-nodes.R
```diff
@@ -287,36 +287,104 @@ protect_tickbox <- function(body, ns) {
...
+#' Protect unescaped square brackets from being escaped
+#'
+#' Commonmark allows both `[unescaped]` and `\[escaped\]` square brackets, but
+#' in the XML representation, it makes no note of which square brackets were 
+#' originally escaped and thus will escape both in the output.
+#'
+#' To fix this, we dive back into the source document and find the brackets that
+#' were originally escaped and wrap those characters in protective tags so that
+#' they do not become escaped when the XML is written. This has the benefit of 
+#' protecting footnotes and reference style links for child documents.
```
maelle, line 300: what are "reference style links for child documents"? could you add a link or example?

R/asis-nodes.R
```diff
@@ -287,36 +287,104 @@ protect_tickbox <- function(body, ns) {
...
+#'
+#' @inheritParams resolve_anchor_links
+#' @param offset the offset in lines of the yaml header defaults to 0,
```
maelle, line 303: maybe clarify?

R/asis-nodes.R
```diff
@@ -287,36 +287,104 @@ protect_tickbox <- function(body, ns) {
...
+#' indicating that the text has the yaml header stripped off. 
+#' @keywords internal
+#' @note This requires documents to be read in with `sourcepos = TRUE`
```
maelle, line 306: maybe specify what read in is here? new object created/other tinkr function used.

R/asis-nodes.R
```diff
@@ -287,36 +287,104 @@ protect_tickbox <- function(body, ns) {
...
+#'
+#' @examples
+#' f <- system.file("extdata", "link-test.md", package = "tinkr")
+#' md <- yarn$new(f, sourcepos = TRUE, unescaped = FALSE)
+#' md$show()
+#' if (requireNamespace("withr")) {
+#' lnks <- withr::with_namespace("tinkr",
```
maelle, line 313: why the withr use here?

R/asis-nodes.R
```diff
@@ -287,36 +287,104 @@ protect_tickbox <- function(body, ns) {
...
+#' @keywords internal
+#' @note This requires documents to be read in with `sourcepos = TRUE`
+#'
+#' @examples
+#' f <- system.file("extdata", "link-test.md", package = "tinkr")
+#' md <- yarn$new(f, sourcepos = TRUE, unescaped = FALSE)
+#' md$show()
+#' if (requireNamespace("withr")) {
+#' lnks <- withr::with_namespace("tinkr", 
+#'   protect_unescaped(md$body, readLines(md$path), offset = length(md$yaml)))
```
maelle, line 314: why couldn't `md` be the argument and all these things the default values?

R/asis-nodes.R
```diff
@@ -287,36 +287,104 @@ protect_tickbox <- function(body, ns) {
...
+#' md$body <- lnks
+#' md$show()
+#' }
+protect_unescaped <- function(body, txt, offset = 0L, ns = md_ns()) {
+  has_sourcepos <- xml2::xml_find_lgl(body, "boolean(.//@sourcepos)")
+  if (!has_sourcepos) {
+    warning("`protect_unescaped()` needs to know the source position. No modification taking place.")
```
maelle, line 321: Hint about how to read in?

R/asis-nodes.R
```diff
@@ -287,36 +287,104 @@ protect_tickbox <- function(body, ns) {
...
 }
 
-protect_footnotes <- function(body, ns = md_ns()) {
-  body <- copy_xml(body)
 
-  feet <- footnote_check(body, ns)
-  if (length(feet) == 0) {
-    return(body)
+find_escaped_squares <- function(txt) {
+  gregexpr("(\\\\\\])|(\\\\\\[)", txt)
```
maelle, line 335: :sweat_smile:

R/asis-nodes.R
```diff
@@ -287,36 +287,104 @@ protect_tickbox <- function(body, ns) {
...
+fix_unescaped_squares <- function(nodes, txt, squares, escapes, offset) {
+  lines  <- get_linestart(nodes) + offset
+  for (i in seq_along(lines)) {
+    this_line <- lines[[i]]
+    this_node <- nodes[[i]]
+    if (!this_line %in% escapes) {
+      # if there are no existing escaped braces here, we need to protect them
+      fix_unescaped(this_node)
+    } else {
+      # if there are escaped braces, there may be situations where we have
```
maelle, line 347: wow good catch

R/to_xml.R
```diff
@@ -10,17 +10,24 @@
...
 #' 
 #' @details This function will take a (R)markdown file, split the yaml header
 #'   from the body, and read in the body through [commonmark::markdown_xml()].
 #'   Any RMarkdown code fences will be parsed to expose the chunk options in
 #'   XML and tickboxes (aka checkboxes) in GitHub-flavored markdown will be
 #'   preserved (both modifications from the commonmark standard). 
 #'
-#'   Math elements 
+#' @note
+#'   Math elements are not protected by default. You can use [protect_math()] to
```
maelle, line 28: should there be an example / test of "apply all possible protections"

tests/testthat/_snaps/anchor-links.md
```diff
@@ -35,12 +35,12 @@
       
       This should also include references that use [standalone][standalone] links and
       footnotes should not be properly parsed and will be considered 'asis' nodes,
-      at least that's what I *believe*\[^footy\]. Maybe this might not pan out \[^but who
-      knows? footnotes are **WEIRD**, man\].
+      at least that's what I *believe*[^footy]. Maybe this might not pan out [^but who
+      knows? footnotes are **WEIRD**, man].
```
maelle, line 39: :joy_cat:

maelle (approved):
Very nice :sparkles:

R/asis-nodes.R
maelle: Thanks! Maybe even an example of what the main and child docs would contain?

R/asis-nodes.R
zkamvar: I was trying to indicate that if a user uses a body from `to_md()` or `yarn$new()` without `sourcepos = TRUE`, then this function can not be used.

R/asis-nodes.R
zkamvar: Because this is an internal function and CRAN did not like my habit of documenting internal functions, so this allows me to access the internal function without using the triple colons, though, I guess I could use `asNamespace("tinkr")$protect_unescaped`

R/to_xml.R
zkamvar: Oh yes absolutely!

We should do a comparison of _everything off_ and then gradually build up to _everything on_. Of course the `yarn` implementation would be much simpler, because it can hide the complexity and allow folks to chain the protections. 

```r
source <- readLines(path)[-(1:3)] # source without yaml

# unescaped protection not possible
A <- to_xml(path, anchor_links = FALSE, unescaped = FALSE)
B <- to_xml(path, anchor_links = TRUE, unescaped = FALSE)

# equivalent to B
A$body <- anchor_links(A$body, source)

# unescaped protection possible
C <- to_xml(path, anchor_links = FALSE, unescaped = FALSE, sourcepos = TRUE)
D <- to_xml(path, anchor_links = TRUE, unescaped = TRUE, sourcepos = TRUE)

# equivalent to D
C$body <- anchor_links(C$body, source)
C$body <- protect_unescaped(C$body, source)
```

I'm realising that all of these are kind of 'off-label' enhancements we offer and I need to write a vignette that spells them out (including what happens with code chunks). The reason `anchor_links` and the `unescaped` are default is that they need the source document in order to process the elements that were hidden by the parser. The reason `protect_math` is not a default is because it is still not fully reliable as it will bork things that look like math (no matter how much I tried to avoid that).

R/asis-nodes.R
zkamvar: I'm thinking that I should remove this since it's not an option in the anchor links function and it requires the same thing.

R/asis-nodes.R
zkamvar: This function does not expect a yarn object, so having `md` be the argument doesn't make sense in this context... and I just realised that I never added a yarn method 🙈

R/asis-nodes.R
zkamvar: This took me a good five minutes to figure out how to get working 😅

NEWS.md
zkamvar: I left it at `unescaped` because I was thinking that we can add in [other commonly escaped characters](https://spec.commonmark.org/0.30/#example-14) later. The only reason why I do not do this now is that I need to make sure the math protection is not adversely affected by this feature (e.g. will having existing `asis` nodes break the search for math delimiters?).

PR #82, opened 2022-12-06
DRAFT in-chunk options support

### LINKED ISSUE #55: Add support for chunk options via knitr 1.35
So, {knitr} had an exciting release this week that [introduced chunk options inside code blocks](https://github.com/yihui/knitr/releases/tag/v1.35) (which supports quarto documents).

It would be a good idea for us to implement this feature, but it may involve borrowing code from {knitr} since the parser not exported (but it could be if we ask nicely!)

One of the good things about this scheme is that we wouldn't have to worry about parsing the chunk options back into the code since the code would be valid no matter what.

here's the roadmap I see:

 - [ ] borrow parser from knitr
 - [ ] add parser to `transform_params()`, with an additional attribute that indicates the params are in the code comments
 - [ ] update `transform_block()` to only write the language if the above above attribute present.

maelle:
Fix #55

### DIFF
--- a/R/to_xml.R
+++ b/R/to_xml.R
@@ -8,21 +8,21 @@
 #'   source position of the file will be included as a "sourcepos" attribute.
 #'   Defaults to `FALSE`.
 #' @param anchor_links if `TRUE` (default), reference-style links with anchors
-#'   (in the style of `[key]: https://example.com/link "title"`) will be 
+#'   (in the style of `[key]: https://example.com/link "title"`) will be
 #'   preserved as best as possible. If this is `FASLE`, the anchors disappear
 #'   and the links will appear as normal links. See [resolve_anchor_links()] for
 #'   details.
 #'
 #' @return A list containing the YAML of the file (yaml)
 #' and its body (body) as XML.
-#' 
+#'
 #' @details This function will take a (R)markdown file, split the yaml header
 #'   from the body, and read in the body through [commonmark::markdown_xml()].
 #'   Any RMarkdown code fences will be parsed to expose the chunk options in
 #'   XML and tickboxes (aka checkboxes) in GitHub-flavored markdown will be
-#'   preserved (both modifications from the commonmark standard). 
+#'   preserved (both modifications from the commonmark standard).
 #'
-#'   Math elements 
+#'   Math elements
 #' @export
 #'
 #' @examples
@@ -47,7 +47,7 @@ to_xml <- function(path, encoding = "UTF-8", sourcepos = FALSE, anchor_links = T
   parse_rmd(body)
   body <- protect_tickbox(body, md_ns())
   if (anchor_links) {
-    body <- resolve_anchor_links(body, splitted_content$body) 
+    body <- resolve_anchor_links(body, splitted_content$body)
   }
 
   list(yaml = yaml,
@@ -76,6 +76,14 @@ transform_block <- function(code_block){
   info <- transform_params(info)
   # This prevents partial code blocks that are still apparently valid: ```{r, }
   info <- info[names(info) != ""]
+  names(info) <- paste0(names(info), "-outchunk")
+
+  code <- strsplit(xml2::xml_text(code_block), "\n")[[1]]
+  inchunk_info <- knitr::partition_chunk("r", code)
+  xml2::xml_text(code_block) <- "code"
+  inchunk_options <- inchunk_info$options
+  names(inchunk_options) <- paste0(names(inchunk_options), "-inchunk")
+  info <- c(info, inchunk_options)
 
   xml2::xml_set_attr(code_block, "info", NULL)
   # preserve the original non-info attributes (e.g. sourcepos)
--- a/inst/extdata/chunk-options.Rmd
+++ b/inst/extdata/chunk-options.Rmd
@@ -1,0 +1,15 @@
+---
+title: "Untitled"
+author: "M. Salmon"
+date: "September 6, 2018"
+output: html_document
+---
+
+## R Markdown
+
+```{r, message=FALSE}
+#| echo = FALSE, fig.width = 10,
+#| fig.cap = "This is a long caption."
+
+plot(cars)
+```
--- a/tests/testthat/test-chunk-options.R
+++ b/tests/testthat/test-chunk-options.R
@@ -1,0 +1,4 @@
+test_that("can parse inside chunk options", {
+  pathchunk <- system.file("extdata", "chunk-options.Rmd", package = "tinkr")
+  tinkr::yarn$new(pathchunk)
+})


R/to_xml.R
```diff
@@ -87,6 +87,19 @@ transform_block <- function(code_block){
   # This prevents partial code blocks that are still apparently valid: ```{r, }
   info <- info[names(info) != ""]
 
+  names(info)[!(names(info) %in% c("language", "name"))] <-
+    paste0(names(info)[!(names(info) %in% c("language", "name"))], "-outchunk")
+
+  code <- strsplit(xml2::xml_text(code_block), "\n")[[1]]
+  inchunk_info <- knitr::partition_chunk(info[["language"]], code)
+  xml2::xml_text(code_block) <- paste0(inchunk_info$code, "\n")
```
maelle, line 95: do not forget this newline :fire:

tests/testthat/_snaps/chunk-options/chunk-options.Rmd
```diff
@@ -0,0 +1,18 @@
...
+date: "September 6, 2018"
+output: html_document
+---
+
+## R Markdown
+
+```{r, message=FALSE}
+#| echo: FALSE
+#| fig.width: 10
+#| fig.cap: This is a long caption.
```
maelle, line 13: maybe that's somewhere where I need to add more YAML-ing as the absence of quote could be problematic?

tests/testthat/_snaps/to_md/to_md-works-for-Rmd.Rmd
```diff
@@ -33,7 +33,7 @@ plot(pressure)
 
 Non-RMarkdown blocks are also considered
 
-```{julia, info=bash}
+```{julia}
```
maelle, line 36: oops

maelle: actually I am not sure why it used to be "info=bash"?

zkamvar: I just found why:

I was trying to test the integrity of the object (e.g. that it was copied before writing instead of transformed in place). I did not consider the fact that I was overwriting a markdown code block within the chunks:

https://github.com/ropensci/tinkr/blob/60d35732c24dcd3cd4761055ec8061b97a55fd43/tests/testthat/test-to_md.R#L98-L110

That being said, I think the behaviour of having the 'info' propagate does actually make sense because it's not unheard of to have custom chunk options.

zkamvar (commented):
Thank you thank you thank you for tackling this! You brought up a good point about what we want the output to be. This PR produces yaml, but we may be able to choose one method and then use [`knitr::convert_chunk_header()`](https://github.com/yihui/knitr/blob/f440712ae6f0546dd8e0354074fb33113cd57318/R/parser.R#L791) to mix and match after that.

What do you think?

R/to_xml.R
```diff
@@ -87,6 +87,19 @@ transform_block <- function(code_block){
   # This prevents partial code blocks that are still apparently valid: ```{r, }
   info <- info[names(info) != ""]
 
+  names(info)[!(names(info) %in% c("language", "name"))] <-
+    paste0(names(info)[!(names(info) %in% c("language", "name"))], "-outchunk")
```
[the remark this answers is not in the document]
maelle: do you mean https://twitter.com/jennybryan/status/1412140590842597385?

R/to_xml.R
maelle: what do you mean? I think it makes sense because the user might change the options via the XML so we need to write the in-chunk options again.

maelle:
or, in the XML, we store the inchunk options as... YAML string? :thinking:

tests/testthat/_snaps/to_md/to_md-works-for-Rmd.Rmd
maelle: I don't get why my code changed this, though.

maelle:
It'd be nice to have a function operating on code chunks that would present chunk options as a tibble, with for each option a row with
- name
- value (list)
- whether it's defined inside or outside the chunk.

Then there'd be a function to carry edits to this tibble to the code chunk XML.

@zkamvar how would that sound? what would this feature be, code chunks as objects?

R/to_xml.R
```diff
@@ -87,6 +87,22 @@ transform_block <- function(code_block){
...
+  outchunk_option <- !(names(info) %in% c("language", "name"))
+  names(info)[outchunk_option] <- paste0(names(info)[outchunk_option], "-outchunk")
+
+  code <- strsplit(xml2::xml_text(code_block), "\n")[[1]]
+  inchunk_info <- knitr::partition_chunk(info[["language"]], code)
+  xml2::xml_text(code_block) <- paste0(inchunk_info$code, "\n")
+  inchunk_options <- inchunk_info$options
+
+  inchunk_options <- if (!is.null(inchunk_options) > 0) {
+    yaml::as.yaml(inchunk_options)
```
maelle, line 99: now it's a bit annoying that there's such different treatment for in- and out- chunk options, it feels a bit wrong maybe.

maelle:
one might first need a method to get all code chunks? :thinking:

R/to_xml.R
zkamvar: That's a good point. I was thinking a situation where the user may want to know how many lines of code they have including options. But you are right. It would be more complex to retain the chunk option formatting in the first place.

R/to_xml.R
```diff
@@ -87,6 +87,22 @@ transform_block <- function(code_block){
   # This prevents partial code blocks that are still apparently valid: ```{r, }
   info <- info[names(info) != ""]
 
+  outchunk_option <- !(names(info) %in% c("language", "name"))
+  names(info)[outchunk_option] <- paste0(names(info)[outchunk_option], "-outchunk")
+
+  code <- strsplit(xml2::xml_text(code_block), "\n")[[1]]
+  inchunk_info <- knitr::partition_chunk(info[["language"]], code)
+  xml2::xml_text(code_block) <- paste0(inchunk_info$code, "\n")
```
zkamvar, line 95: Also, I was wondering why the Julia example was being truncated and it's because of this nasty property of the text for code chunks: the input needs to be a single length vector

```suggestion
  xml2::xml_text(code_block) <- paste0(paste(inchunk_info$code, collapse = "\n"), "\n")
```

R/to_xml.R
zkamvar: Safety first! 🥽 

```suggestion
    yaml::as.yaml(inchunk_options, eval.expr = FALSE)
```

zkamvar: This will help parse expressions

```suggestion
    yaml::as.yaml(inchunk_options,
      handlers = list(expression = function(x) paste("!expr: ", paste(as.character(x), sep = "; ")))
```

zkamvar:
> one might first need a method to get all code chunks? :thinking:

This is a good point. We've been able to get away with parsing the code chunks because they've always been generally linear and not very complex. With the new nested complexity that YAML introduces, It would definitely make sense for us to have options to store and access the chunk options consistently.

I'm wondering if we should just be opinionated on how code chunks are treated on output (e.g. users can choose to use curly, hashpipe-list, or hashpipe-yaml, but no mixing).

R/to_xml.R
maelle: Oh, good catch!!

R/to_xml.R
maelle: @zkamvar could you please explain this?

zkamvar: This is to address expressions that originate from YAML, but I'm now realising that I'm actually having a hard time getting a good example, so I have to rethink this. 

``` r
(x <- list(A = 1, B = "two", C = str2expression("C <- 1 + 2\nC")))
#> $A
#> [1] 1
#> 
#> $B
#> [1] "two"
#> 
#> $C
#> expression(C <- 1 + 2, C)
eval(x$C) # C evaluates to 3
#> [1] 3
try(yaml::as.yaml(x)) # conversion fails because C is an expression
#> Error in yaml::as.yaml(x) : Unknown emitter error
expr_handler <- function(x) paste("!expr:", paste(as.character(x), collapse = "; "))
yaml::as.yaml(x, handlers = list(expression = expr_handler))
#> [1] "A: 1.0\nB: two\nC: '!expr: C <- 1 + 2; C'\n"
```

<sup>Created on 2023-01-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

tests/testthat/_snaps/to_md/to_md-works-for-Rmd.Rmd
maelle: @zkamvar why was there originally the "info=bash"?

R/to_xml.R
maelle: should this go into a feature where we transform code chunks info into a tibble? I mean adding a column with number of code lines.

tests/testthat/_snaps/chunk-options/chunk-options.Rmd
maelle: I think it is solved? see the julia chunk below

maelle:
@zkamvar what would be a good target for this PR? Happy to do more work on it / follow-up PRs :grin:

zkamvar:
> @zkamvar what would be a good target for this PR? Happy to do more work on it / follow-up PRs grin

Ugh, I'm so sorry. We were doing this when things got really hectic for me and I had to drop it. I'm going to come back to this probably after May (that's when we transition all of our lessons)

maelle:
No worry and good luck with the lessons transition! :rocket:

PR #84, opened 2022-12-09
update check to no longer use ubuntu 18.04

zkamvar:
This updates the check-full workflow while retaining the cron job that would check every week.


zkamvar:
Note: I updated the branch protection settings to no longer require 18.04. I will add it to require ubuntu-latest when I merge this.

Issue #85, opened 2023-01-25
protect text inside of anchor link keys


zkamvar:
Whilst planning the transition for Carpentries lessons, I ran into a situation where a page with anchor links were being mangled: 

``` r
txt <- "**Maintainer(s):**


* [Sarah Brown][brown_sarah]: @brownsarahm
* [Tim Dennis][dennis_tim]: @jt14den
* [David Perez-Suarez][perez-suarez_david]: @dpshelio
* [Nathaniel Porter][porter-nathaniel]: @ndporter
* [Jon Wheeler][wheeler_jon]: @jonathanwheeler01
* [Karen Word][word_karen]: @karenword  


[dc-site]: http://datacarpentry.org
[lesson-example]: https://carpentries.github.io/lesson-example
[swc-site]: http://software-carpentry.org
[lc-site]: https://librarycarpentry.org
[koch_christina]: https://carpentries.org/instructors/
[brown_sarah]: https://carpentries.org/instructors/ 
[dennis_tim]: https://carpentries.org/instructors/
[perez-suarez_david]: https://carpentries.org/instructors/
[porter-nathaniel]: https://carpentries.org/instructors/
[wheeler_jon]: https://carpentries.org/instructors/
[word_karen]: https://carpentries.org/team/"
tmp <- tempfile()
writeLines(txt, tmp)
tinkr::yarn$new(tmp)$show()
#> **Maintainer(s):**
#> 
#> - [Sarah Brown][wheeler_jon]: @brownsarahm
#> - [Tim Dennis][wheeler_jon]: @jt14den
#> - [David Perez-Suarez][wheeler_jon]: @dpshelio
#> - [Nathaniel Porter][wheeler_jon]: @ndporter
#> - [Jon Wheeler][wheeler_jon]: @jonathanwheeler01
#> - [Karen Word][word_karen]: @karenword
#> 
#> [wheeler\_jon]: https://carpentries.org/instructors/
#> [word\_karen]: https://carpentries.org/team/
```

<sup>Created on 2023-01-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

There are actually two issues here:

1. anchor links with duplicate links are coalesced into the last one
2. the underscore character is not protected in the anchor link definition

zkamvar:
I seem to have forgotten to include relevant context:

https://github.com/carpentries/lesson-transition/issues/15

zkamvar:
[...]

PR #86, opened 2023-01-25
don't escape relative link anchors

zkamvar:
This will address #85 by not escaping relative link anchors if they have escapable characters.

Since these are not rendred by markdown to HTML, there is no reason to escape them, thus this pattern will now work:

```markdown
here is a [linke that we use an underscore in][underscored_link]. It will not
be escaped when we roundtrip with {tinkr}!

[underscored_link]: https://github.com/ropensci/tinkr/issues/85
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@
   unescaped square braces to remain unescaped (#78). 
 * New `protect_curly()` function will add a `curly='true'` attribute to text wrapped in curly braces ('{', '}') to allow parsing of the XML for sending to external APIs. This function will also parse alt text and place it in an attribute.
 * New `$protect_curly()` method implements `protect_curly()` on yarn objects
-* relative link anchor keys will no longer have characters escaped (#85)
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -6,6 +6,7 @@
   unescaped square braces to remain unescaped (#78). 
 * New `protect_curly()` function will add a `curly='true'` attribute to text wrapped in curly braces ('{', '}') to allow parsing of the XML for sending to external APIs. This function will also parse alt text and place it in an attribute.
 * New `$protect_curly()` method implements `protect_curly()` on yarn objects
+* relative link anchor keys will no longer have characters escaped (#85)
```
maelle, line 9:
```suggestion
* Relative link anchor keys will no longer have characters escaped (#85).
```

Peak nitpickiness :sweat_smile:

maelle (approved):
:rocket: especially as this touches the xsl file!

PR #88, opened 2023-01-26
refactor: rm dependency on stringr

### LINKED ISSUE #87: remove stringr dependency?
With {knitr} no longer requiring {stringr}, it makes sense that we should probably also drop the dependency: https://github.com/yihui/knitr/releases/tag/v1.42

maelle:
Fix #87

### DIFF
--- a/R/utils-string.R
+++ b/R/utils-string.R
@@ -1,20 +1,20 @@
 # Replacements for stringr functions
 # to preserve the argument order + nice name
 
-str_remove_all <- function(string, pattern) {
-  gsub(pattern, "", string)
-}
-
-str_remove <- function(string, pattern) {
-  sub(pattern, "", string)
-}
-
 str_replace <- function(string, pattern, replacement) {
   sub(pattern, replacement, string)
 }
 
 str_replace_all <- function(string, pattern, replacement) {
   gsub(pattern, replacement, string)
+}
+
+str_remove <- function(string, pattern) {
+  str_replace(string, pattern, "")
+}
+
+str_remove_all <- function(string, pattern) {
+  str_replace_all(string, pattern, "")
 }
 
 str_detect <- function(string, pattern) {
--- a/R/utils-string.R
+++ b/R/utils-string.R
@@ -10,11 +10,11 @@ str_remove <- function(string, pattern) {
 }
 
 str_replace <- function(string, pattern, replacement) {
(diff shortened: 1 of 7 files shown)


tests/testthat/_snaps/asis-nodes.md
```diff
@@ -3,7 +3,7 @@
     Inline math delimiters are not balanced.
     
     HINT: If you are writing BASIC code, make sure you wrap variable
-          names and code in backtics like so: `INKEY$`. 
+          names and code in backtics like so: `INKEY$`.
```
maelle, line 6: the hint in the code also doesn't have this whitespace :thinking:

zkamvar: No worries! I think your editor trimmed the trailing whitespace of [asis-nodes.R#L13](https://github.com/ropensci/tinkr/pull/88/files#diff-60bef5af43a1bd541299ce9bceeb226afb23deaa5e637b145c85e05ffd552b0cL13-R13) when you updated the stringr function.

R/utils-string.R
```diff
@@ -0,0 +1,26 @@
+# Replacements for stringr functions
+# to preserve the argument order + nice name
```
zkamvar, line 2: I absolutely love this solution! 💯

R/utils-string.R
```diff
@@ -0,0 +1,26 @@
...
+
+str_replace_all <- function(string, pattern, replacement) {
+  gsub(pattern, replacement, string)
+}
+
+str_remove <- function(string, pattern) {
+  str_replace(string, pattern, "")
+}
+
+str_remove_all <- function(string, pattern) {
```
zkamvar, line 16: We don't use this function, but I'm okay keeping it

zkamvar (approved):
YESSSSSS! :tada: 

Why do we need {stringr} when we have `tinkr::yarn` :yarn: 😝?

I was slightly dreading the difficulty of this, but the solution you came up with is spot on and really straightforward!

I have one request: please add a NEWS item 🙏🏼

maelle:
> I was slightly dreading the difficulty of this, but the solution you came up with is spot on and really straightforward!

The commit history furthermore shows how easy it was for me. :zany_face: 

> I have one request: please add a NEWS item 🙏🏼

Done!

PR #93, opened 2023-03-17
If no yaml, avoid prepending with gratuitous `\n`

jennybc:
This PR adapts tinkr for forward compatibility with glue. I will release glue no sooner than 2 weeks from now, so on March 31 (or likely a bit later).

In the next version of glue, `glue::glue_collapse()` will never return `character()` but instead will return `""` for empty inputs. For more detail see https://github.com/tidyverse/glue/pull/295.

In my hands, this PR makes tinkr work as intended with released glue and dev glue.

### DIFF
--- a/R/to_md.R
+++ b/R/to_md.R
@@ -61,7 +61,9 @@ to_md <- function(yaml_xml_list, path = NULL, stylesheet_path = stylesheet()){
 transform_to_md <- function(body, yaml, stylesheet) {
   body <- xslt::xml_xslt(body, stylesheet = stylesheet)
 
-  yaml <- glue::glue_collapse(yaml, sep = "\n")
+  if (length(yaml) > 0) {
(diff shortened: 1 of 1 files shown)


zkamvar:
Thank you! I will admit, when I saw that you had forked the repo, I was kind of excited because "OMG JB noticed!" 

I've tested this locally with the dev version of glue and can confirm that it works. I'm going to merge. 

Would you like to be added to the DESCRIPTION as a contributor?

I will definitely get this submitted to CRAN if not today then next week by Wednesday.

jennybc:
> Would you like to be added to the DESCRIPTION as a contributor?

No thanks, this is just me fixing a small thing we're going to break 😅 

> I will definitely get this submitted to CRAN if not today then next week by Wednesday.

Thanks!

Issue #92, opened 2023-02-23
footnote bug?


maelle:
``` r
lines <- c("lala[^1]", "", "[^1]: pof")

temp_file <- withr::local_tempfile()
brio::write_lines(lines, temp_file)

tinkr::yarn$new(temp_file)$write(temp_file)

brio::read_lines(temp_file)
#> [1] "lala[^1][^1]" ""             "[^1]: pof"    ""             ""            
#> [6] ""

# second one
lines <- c("lala[^1] blabla", "", "[^1]: pof")

temp_file <- withr::local_tempfile()
brio::write_lines(lines, temp_file)

tinkr::yarn$new(temp_file)$write(temp_file)

brio::read_lines(temp_file)
#> [1] "lala[^1][^1] blabla" ""                    "[^1]: pof"          
#> [4] ""                    ""                    ""
```

<sup>Created on 2023-02-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

zkamvar:
OMG why didn't I see this!?

maelle:
Maybe because you write clear text that doesn't need footnotes?!

zkamvar:
> Maybe because you write clear text that doesn't need footnotes?!

`*wheezing*`

Funny enough, the most recent update of {commonmark} has "footnotes" support, but it's a bit strange:

``` r
txt <- c("a statement[^1][^2]\n", "[^1]: this is true", "[^2]: this is false")
commonmark::markdown_xml(txt, footnotes = TRUE) |> writeLines()
#> <?xml version="1.0" encoding="UTF-8"?>
#> <!DOCTYPE document SYSTEM "CommonMark.dtd">
#> <document xmlns="http://commonmark.org/xml/1.0">
#>   <paragraph>
#>     <text xml:space="preserve">a statement</text>
#>     <<unknown> />
#>     <<unknown> />
#>   </paragraph>
#>   <<unknown>>
#>     <paragraph>
#>       <text xml:space="preserve">this is true</text>
#>     </paragraph>
#>   </<unknown>>
#>   <<unknown>>
#>     <paragraph>
#>       <text xml:space="preserve">this is false</text>
#>     </paragraph>
#>   </<unknown>>
#> </document>
```

<sup>Created on 2023-03-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

zkamvar:
OH I think I know what's going on! {tinkr} sees this footnote as a relative link:

[...]

PR #95, opened 2023-04-14
Footnotes should not be escaped

maelle:
@zkamvar this is a follow-up to #92 

The difference is that the footnote text is more than one word.

I'm still exploring. :thinking:


maelle:
I guess it's linked to the commonmark issue you posted :sweat_smile:

zkamvar:
It works when you use `sourcepos = TRUE` (that's how the protect unescaped works), but it's weird that the original example works and this one doesn't

PR #102, opened 2024-04-05
test: remove top-level code in test-anchor-links.R

### LINKED ISSUE #83: Rm code outside of test_that()
Using helper functions/helper files/duplication.

maelle:
Fix #83

### DIFF
--- a/.github/workflows/check-full.yaml
+++ b/.github/workflows/check-full.yaml
@@ -6,9 +6,9 @@
 # usethis::use_github_action("check-standard") will install it.
 on:
   push:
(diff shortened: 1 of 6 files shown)


zkamvar (approved):
Thank you for taking care of this and for labeling your commits so well!

I'm happy for you to merge this. 

I also noticed that the R-CMD-check workflow has not been running because it was auto-disabled by GitHub a while ago and I never re-enabled it. If you want to push an empty commit to kickstart it again, that would be good, otherwise merge as you wish :)

PR #112, opened 2024-05-10
filter out special control characters

### LINKED ISSUE #96: filter out / replace special control characters
I'm running into a fun situation where someone added a `\v` control character to the end of a list and I'm getting a `PCDATA invalid Char value 11` error.

There is an answer in SO that [describes how to detect and replace the bad character with a blank](https://stackoverflow.com/questions/26408727/pcdata-invalid-char-in-r), so we can take that and add it to the pre-processing step.

NOTE: Since I can hear Nero fiddling on the outskirts of SO, here's the solution:

```r
    illegal <- "[^\u0009\u000a\u000d\u0020-\uD7FF\uE000-\uFFFD]" 
    
    utf8_for_xml <- function(x) {
        
        return(gsub(illegal, "", x))
        
        }

    string_formatted <- utf8_for_xml(string)
```

I believe this is the correct way to do it since these characters shouldn't exist in the document in the first place. That being said, it's a bit frustrating to find out what these characters actually do, but I've found these resource:

- https://en.wikipedia.org/wiki/Control_character
- https://en.wikipedia.org/wiki/Valid_characters_in_XML#XML_1.1

zkamvar:
this will fix #96

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 * Inline math with single characters will no longer cause an error (issue: #101,
   fix: #103, @maelle)
+* Special control characters are now filtered out before processing XML (issue:
+  #96, fix: #111, @zkamvar)
 
 ## MISC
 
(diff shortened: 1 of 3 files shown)


zkamvar:
I am going to go ahead and merge this without review as it's a fairly straightforward fix that will only improve user experience.

Issue #121, opened 2024-10-10
how do we protect money dollar sign from math protection


maelle:
:wave: @zkamvar I am realizing I do not know how to fix the problem below without resorting to backticks:

``` r
lines <- "With this $340k granta,"
file <- withr::local_tempfile()
brio::write_lines(lines, file)

tinkr::yarn$new(file)$protect_math()
#> Error: Inline math delimiters are not balanced.
#> 
#> HINT: If you are writing BASIC code, make sure you wrap variable
#>       names and code in backtics like so: `INKEY$`.
#> 
#> Below are the pairs that were found:
#>                   start...end
#>                   -----...---
#> With this $340k granta,...
```

<sup>Created on 2024-10-10 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Any hint?

zkamvar:
oof that's not great and is definitely a bug.

zkamvar:
Ugh.... I realize now that all of the examples we have are balanced -_-. I'm not sure why I never noticed this before: 

https://github.com/ropensci/tinkr/blob/8a57235f7072dbcb7eb22516290c58b692a9c015/inst/extdata/basic-math.md?plain=1#L7-L14

zkamvar:
I think the solution may be in here:

https://github.com/ropensci/tinkr/blob/8a57235f7072dbcb7eb22516290c58b692a9c015/R/asis-nodes.R#L143-L149

If we check that `lh` is zero, then we can exit early.

zkamvar:
As noted in https://github.com/ropensci/tinkr/pull/122#issue-2580161441, this will still be a problem if there is broken math that needs fixing:

```markdown
This is an example with a bunch of broken math.

## Broken math mixed with currency. 

I bet you $5 that 

$P \ne
NP$

$P 
\eq NP$

I bet you $10 that... no,
I bet you $20 that 

$P \gt
NP$

$P 
\lt NP$

Give me $5.

## broken math on the same line

$P \eq
NP$ and $P \ne
NP$
```

We use regex to detect the broken math and isolate the text nodes that contain the broken math. Since broken math is generally on adjacent lines, we assume that these nodes are next to each other where the odd nodes are the opening sign and the even nodes are the corresponding closing signs. 

Currency dollars throw this off because they create an offset in the opening signs and need to be addressed, but I get a feeling it will need to be a sort of recursive function that walks down the nodes, checks for pairs, discarding unpaired dollars along the way so that the zipper works.

PR #123, opened 2024-10-16
add recursive function to find currency dollars

zkamvar:
This adds a stronger solution to #121.

It does a second check to see if there are more opening dollar signs than
closing and then passes those through a function that walks the list of broken
math elements and detects the opening elements that aren't paired with closing
elements (by confirming that the pairs are adjacent). The opening elements that
aren't paired are excluded from the list and then if the list is still not
balanced an error is thrown. 

Update 2025-10-06: I worked on this a bit yesterday and found that I might be able to actually detect the differences between Mal-formed math and stray currency notation in https://github.com/ropensci/tinkr/pull/123/commits/710224353e4c2cdd3ed65134b1f6b1f1d4ae09ad. 

With this PR, we are being greedy with math protection, as we are throwing out math elements that do not conform to our expectations. However, the way we are doing it does not take into account the fact that a node could be marked as ambiguous (has no start and no end markers). I think when we take this into account, we can properly detect broken math elements and throw a warning for them instead of throwing an error.


R/asis-nodes.R
```diff
@@ -129,7 +129,7 @@ protect_inline_math <- function(body, ns) {
   }
 
   # protect math that is broken across lines or markdown elements
-  if (length(bmath)) {
```
maelle, line 132: maybe use an explaining variable although that is already clear

R/asis-nodes.R
```diff
@@ -140,15 +140,26 @@ protect_inline_math <- function(body, ns) {
...
-    # If the lengths of the beginning and ending tags don't match, we throw
-    # an error.
     le <- length(bmath[endless])
     lh <- length(bmath[headless])
     # 2024-10-10: if the number of headless tags is zero, then we are dealing
     # with currency. See issue #121 
     if (lh == 0) {
       return(copy_xml(body))
     }
+    # 2024-10-15: if the number of headless tags is _less_ than the number of
```
maelle, line 150:
```suggestion
    # 2024-10-15: if the number of startless tags is _less_ than the number of
```

maelle: the variable is actually headless, but I wonder if it might be confusing

R/asis-nodes.R
```diff
@@ -163,6 +174,58 @@ protect_inline_math <- function(body, ns) {
...
+  } else {
+    # CASE 2: MISMATCHED PAIRS --------------------------------
+    # When the pairs do not match, it's not likely broken math, 
+    # so we toss it and move to the next element. 
+    result <- FALSE
+    idx <- -1
+  }
+  # return the result and iterate over the rest of the vector
+  return(c(result, toss_broken_teeth(endless[idx], headless[idx])))
+}
```
maelle, line 227: should we use devtag https://github.com/moodymudskipper/devtag

and have examples

man/rmd-fragments/format-latex.Rmd
```diff
@@ -19,31 +19,17 @@ Note, however, that there are a few caveats for this:
...
+      ```{r}
+      # no separator, math is detected at the beginning of the line.
+      ex <- textConnection("$5 for a $\\pi$")
+      tinkr::yarn$new(ex)$protect_math()$get_protected()
+      # With an emph separator
+      ex <- textConnection("$5 for _any_ $\\pi$")
+      tinkr::yarn$new(ex)$protect_math()$get_protected()
+      ```
+   2. Postfix dollar signs (e.g. using BASIC code demostrations) should be
+      protected with backtics, or it will cause an error.
```
maelle, line 34:
```suggestion
      protected with backticks, or it will cause an error.
```

man/rmd-fragments/format-latex.Rmd
```diff
@@ -19,31 +19,17 @@ Note, however, that there are a few caveats for this:
...
+      the math unless a markdown element separates them:
+      ```{r}
+      # no separator, math is detected at the beginning of the line.
+      ex <- textConnection("$5 for a $\\pi$")
+      tinkr::yarn$new(ex)$protect_math()$get_protected()
+      # With an emph separator
+      ex <- textConnection("$5 for _any_ $\\pi$")
+      tinkr::yarn$new(ex)$protect_math()$get_protected()
+      ```
+   2. Postfix dollar signs (e.g. using BASIC code demostrations) should be
```
maelle, line 33:
```suggestion
   2. Postfix dollar signs (e.g. using BASIC code demonstrations) should be
```

maelle (commented):
thank you!! another solution: use € not $ :smile_cat:

R/asis-nodes.R
zkamvar: oh good point!

R/asis-nodes.R
zkamvar: I think you have a good point and will change the variable name.

R/asis-nodes.R
zkamvar: That's a good point. I had avoided it because I though "I don't want to learn a new thing" but then I realized that this will definitely help us in the end (especially if it means that we get local docs that CRAN does not care or know about 😈)

man/rmd-fragments/format-latex.Rmd
zkamvar: I have figured out how to turn on spellcheck in vim, so I can now catch silly things like this! 😅

R/asis-nodes.R
zkamvar: done!

man/rmd-fragments/format-latex.Rmd
maelle: nice!!

PR #127, opened 2024-12-16
feat: frontmatter

### LINKED ISSUE #126: expand tinkr to reading metadata in toml?
If the first line is "---" (or more I guess) it is YAML, if it is "+++" it is TOML.

The package wouldn't need to parse TOML, just to keep it as text to not break it.

Happy to make a PR.

@yabellini

maelle:
Fix #126


R/utils.R
```diff
@@ -24,36 +24,45 @@ unbalanced_math_error <- function(bmath, endless, headless, le, lh) {
...
-    yaml_load(res$yaml[-c(1, n)])
-  }
+
   res
 }
 
-# https://github.com/rstudio/blogdown/blob/9c7f7db5f11a481e1606031e88142b4a96139cce/R/utils.R#L407
-# anotate seq type values because both single value and list values are
-# converted to vector by default
-yaml_load = function(x) yaml::yaml.load(
```
maelle, line 45: Was this used anywhere? Not in the tests at least :sweat_smile:

R/utils.R
```diff
@@ -24,36 +24,45 @@ unbalanced_math_error <- function(bmath, endless, headless, le, lh) {
...
+    return(yamlish)
+  }
+
+  tomlish <- split_thing_body(x, '^\\+\\+\\+\\s*$')
+  if (length(tomlish[["frontmatter"]]) > 0) {
+    tomlish$frontmatter_format <- "TOML"
+    return(tomlish)
+  }
+
+  jsonish <- split_thing_body(x, '^\\{\\s*$', '\\}$')
```
zkamvar, line 53: I was running into a false positive with a test in pegboard that had a chunk at the top instead of front matter. The second pattern was returning all the lines that had a closing `}`

I think this would help with that? 

```suggestion
  jsonish <- split_thing_body(x, '^\\{\\s*$', '^[^{]*\\}$')
```

zkamvar (changes requested):
Thank you! Sorry I haven't had time to get to this... I borked my home computer last week and haven't been able to get to this -_-

This looks good and pretty much is exactly what I was expecting! The only thing that is blocking is the regex for the JSON header was swallowing chunks at the beginning of documents.

Note that this will require some coordination with the carpentries to release because this will cause a lot of warnings (try testing pegboard with this release and you get over 70 warnings)

maelle:
@zkamvar the warnings you see in pegboard are the deprecations, correct?

maelle:
I have the impression we're getting the deprecation warning even when not accessing YAML actually. Is there another example of an R6 class deprecating a class that you know of and that I could look at?

maelle:
Forgot to write, thank you for the review!

zkamvar:
> @zkamvar the warnings you see in pegboard are the deprecations, correct?

Yes, those are the deprecation warnings. This is because pegboard's `Episode` class inherits `yarn` and was accessing the `$yaml` field directly.

maelle:
Ok so the plan would be for me to:

- finish this PR (is the example I added correct?)
- make a PR to pegboard
- make a PR to babeldown
- male a PR to aeolus if needed
- make a PR to https://github.com/mpadge/altdoc if needed

Correct?

zkamvar:
Oh hey! Sorry I haven't gotten to this... I've been busy af 😞 

That breakdown makes sense, but I would like to warn @froggleston before this happens since it will cause pegboard tests to fail (though pegboard itself will still work) and throw a lot of innocuous warnings in The Carpentries build infrastructure.

zkamvar:
This is my own dang fault for not releasing to CRAN sooner -_-

froggleston:
Thanks for the heads up @zkamvar ! Should I wait for any PRs to be raised against pegboard, or is there something I can do ahead of time my end?

maelle:
@zkamvar just checking whether you're waiting on any action on my side (this is not to put any pressure on you, this is a genuine question! it's fine if this PR takes time but I don't want to be the speed limiting factor because of forgetfulness :sweat_smile: )

zkamvar:
I think I'm going to try to make a PR to pegboard this weekend and then merge on Monday.

zkamvar:
I've turned off the "delete head branches" for this PR to go through and will merge it.

@froggleston, I'm going to merge this PR and I've opened https://github.com/carpentries/pegboard/pull/163 that will fix whatever warnings pop up for the carpentries users when this gets pushed to the r-universe. I would suggest to merge that PR in before the end of today to avoid too much maintainer confusion. 

Sorry about all the faff and the lateness of all this! I made the foolish mistake to update my OS a few weeks ago and only just now got it back up and running 🙈

froggleston:
Thanks @zkamvar ! Do I need to do a pegboard release?

zkamvar:
> Thanks @zkamvar ! Do I need to do a pegboard release?

Yes

froggleston:
OK, bumped to 0.7.8

PR #131, opened 2025-03-07
feat: protect fenced divs

### LINKED ISSUE #98: protect_triplecolons?
Would it be in scope for tinkr to have a function for protecting triple colons such as

```md
::: footer
[You are sneaky!]{style="color: transparent;"}
:::
```

In babeldown, because they become text, they're translated. I might add something in babeldown.

maelle:
Fix #98 

I am not sure I understand everything that is done in `protect_curly` so this might be too simplified.


NEWS.md
```diff
@@ -28,7 +28,10 @@
   A new field called `frontmatter_format` ("YAML", "TOML", or "JSON")
   contains the format of the frontmatter.
   (issue: #126, @maelle)
-  
+  * New `protect_fences()` function will add a `fence='true'` attribute to fences of Pandoc fenced divs to allow parsing of the XML for sending to
+  external APIs. 
+* New `$protect_fences()` method implements `protect_fences()` on yarn objects
```
maelle, line 33: should it be called `protect_fence()`? I find the plural more natural but the other protection functions use the singular.

R/class-yarn.R
```diff
@@ -54,9 +60,9 @@ yarn <- R6::R6Class("yarn",
       self$frontmatter <- xml$frontmatter
       self$frontmatter_format <- xml$frontmatter_format
       self$body <- xml$body
-      self$ns   <- tinkr::md_ns()
```
maelle, line 57: ouch I think Air did this

zkamvar: Air?

maelle: https://posit-dev.github.io/air/

maelle:
PR in babeldown for info https://github.com/ropensci-review-tools/babeldown/pull/81

NEWS.md
zkamvar: I think the plural makes sense partially because `protect_curlies()` sounds slightly disgusting to my USAian ears 🤮

zkamvar (commented):
This looks good! There is one bit of code that we can get rid of without consequences, which might also mean that we can just run `find_fences` and then add the attribute without having to copy the nodes. 


Sorry for not getting to this until three weeks later -_- It's been a busy few months and my mental health hasn't been the greatest.

zkamvar:
Thank yoU!

PR #141, opened 2025-09-28
use air formatting

zkamvar:
This formats tinkr with air, which also means that all the other PRs will need to be updated to account for these changes, but will make things easier to work with in the future.


inst/scripts/xml_table.R
```diff
@@ -1,13 +1,14 @@
-input <- c("| a  | b  |  c  |  d |", "| :- | -- | :-: | -: |", "| l  | n  |  c  |  r |"
```
zkamvar, line 1: funny enough, air made this _better_

maelle (commented):
Thank you!!

maelle (approved):
thank you!!

PR #144, opened 2025-10-04
release 0.3.0

zkamvar:


zkamvar:
I apparently cannot review my own PRs... and of course, this pull request was supposed to be merged before #147 🤦🏽

maelle (commented):
thank you! So do you need to undo that previous commit?

zkamvar:
no I don't think we do. The record of the commit exists and the tag is in the right place, so it should be fine

PR #145, opened 2025-10-07
Update add_md.R

### LINKED ISSUE #4: tables are not preserved

monikavila:
Change glue::glue. con cli

### DIFF
--- a/R/add_md.R
+++ b/R/add_md.R
@@ -55,7 +55,7 @@ shove_nodes_in <- function(body, new, nodes, where = "after", space = TRUE) {
     xpath <- NULL
   }
   if (length(nodes) == 0) {
-    msg <- glue::glue("No nodes matched the expression {sQuote(xpath)}")
+    msg <-  cli_text("No nodes matched the expression {sQuote(xpath)}")
     rlang::abort(msg, class = "insert-md-xpath")
   }
   if (!inherits(nodes, c("xml_node", "xml_nodeset"))) {


monikavila:
Changed glue::glue by cli::cli_text()

R/add_md.R
```diff
@@ -55,7 +55,7 @@ shove_nodes_in <- function(body, new, nodes, where = "after", space = TRUE) {
     xpath <- NULL
   }
   if (length(nodes) == 0) {
-    msg <- glue::glue("No nodes matched the expression {sQuote(xpath)}")
+    msg <-  cli_text("No nodes matched the expression {sQuote(xpath)}")
```
zkamvar, line 58: There are two things here that can be addressed:  

1. Adding `cli::` to the beginning of the call lets R know where to fetch the function from (we _could_ import this from the {cli} package, but we've made the choice not to because it makes it more clear where the function is coming from).
2. using `cli::format_error()` instead of `cli::cli_text()`. The `cli::format_` functions are designed specifically for `rlang::abort()` and similar: https://cli.r-lib.org/reference/format_error.html


```suggestion
    msg <-  cli::format_error("No nodes matched the expression {sQuote(xpath)}")
```

zkamvar (changes requested):
👋🏽 Welcome and thank you for the contribution! If I am correct, this is a contribution towards #142?

I have some changes to request. The first is in my comment below. You can [use the GitHub web interface to accept the changes directly](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) or you can make them locally and push them up if you wish. 

The second change is to do add your name to the list of contributors in the DESCRIPTION file (you can use [`usethis::use_author()`](https://usethis.r-lib.org/reference/use_author.html), bump the version to development (you can use [`usethis::use_version()`](https://usethis.r-lib.org/reference/use_version.html) to bump the version), and add an item to the NEWS.md file that describes the change under a `MISC` heading.

PR #150, opened 2026-05-02
simplify math delimiter errors

zkamvar:
I noticed that using `cli::cli_format()` was doing funky things to this
error message like adding extra quotes. Also, since we already manually
calculate the column widths, I figure we can keep them.

When going from rlang::abort() to cli, the function to use is
`cli::format_error()` and it will auto-parse all the bracketed
expressions.


tests/testthat/_snaps/asis-nodes.md
```diff
@@ -4,15 +4,14 @@
       me$protect_math()
     Condition
       Error in `unbalanced_math_error()`:
-      ! Inline math delimiters are not balanced.
-      HINT: If you are writing BASIC code, make sure you wrap variable
-                names and code in backtics like so: {.code INKEY$}
+      ! x Inline math delimiters are not balanced.
```
maelle, line 7: Why have both ! and x?

maelle (approved):
Thank you 🚀

tests/testthat/_snaps/asis-nodes.md
zkamvar: oh good point! That can be changed!

PR #156, opened 2026-05-11
chore: replace one loop with a map

### LINKED ISSUE #50: replace use of for loops with purrr walk/walk2?

maelle:
Fix #50

I don't find `reduce()` easy, I use through reading https://masalmon.eu/2023/07/26/reduce/ + trial and error.

This is a 100% manual PR. :sweat_smile:

### DIFF
--- a/R/asis-nodes.R
+++ b/R/asis-nodes.R
@@ -436,27 +436,33 @@ fix_unescaped_squares <- function(nodes, txt) {
   # indicator of which lines have escaped square braces
   escapes <- which(vapply(squares, sum, integer(1)) > 0L)
   lines <- get_linestart(nodes)
-  for (i in seq_along(lines)) {
-    this_line <- lines[[i]]
-    this_node <- nodes[[i]]
-    if (!this_line %in% escapes) {
-      # if there are no existing escaped braces here, we need to protect them
-      fix_unescaped(this_node)
-    } else {
-      # if there are escaped braces, there may be situations where we have
-      # escaped and unescaped braces on the same line (for example a link and
-      # an example of a link). This will tell us if the node we are handling
-      # contain the characters we need to escape (markup splits the nodes).
-      start <- get_colstart(this_node)
-      end <- get_colend(this_node)
-      escape_sequence <- squares[[this_line]]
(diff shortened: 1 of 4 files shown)


R/asis-nodes.R
```diff
@@ -298,9 +297,10 @@ protect_tickbox <- function(body, ns) {
   char <- sub("(\\[.\\])", "\\1</text><text>", char, perl = TRUE)
   new_nodes <- purrr::map(char, make_text_nodes)
   # since we split up the nodes, we have to do this node by node
-  for (i in seq(new_nodes)) {
```
maelle, line 301: an advantage is that we could use https://purrr.tidyverse.org/reference/in_parallel.html but I'm not sure it's worth it?

R/asis-nodes.R
```diff
@@ -437,29 +437,35 @@ fix_unescaped_squares <- function(nodes, txt) {
...
   invisible()
 }
 
+fix_single_unescaped_square <- function(i, nodes, lines, squares, escapes) {
+  this_line <- lines[[i]]
+  this_node <- nodes[[i]]
+  if (!this_line %in% escapes) {
+    # if there are no existing escaped braces here, we need to protect them
+    fix_unescaped(this_node)
+    return(TRUE)
```
maelle, line 454: I was unsure of how to get an early return here.

R/attr-nodes.R
```diff
@@ -6,44 +6,51 @@ find_curly <- function(body, ns) {
...
   curlies <- regmatches(char, gregexpr("\\{.*?\\}", char))[[1]]
-  for (curl in curlies) {
-    attributes <- "curly='true'"
 
-    alt_fragment <- regmatches(curl, gregexpr("alt=(['\"]).*?\\1", curl))[[1]]
-    if (length(alt_fragment) > 0) {
-      alt_text <- sub("^alt=", "", alt_fragment)
-      attributes <- sprintf("%s alt=%s", attributes, alt_text)
-    }
+  purrr::reduce(curlies, absorb_curly, .init = char) |>
```
maelle, line 35: this inspired #155

